Issue #16, opened 2016-03-09
Implement to express


ReverseFlash28:
How can I use this together with express?

kvz:
I think the goal of this project is to have a stand-alone server of tus, although it has been some time that work was done here, did anything change on your end @bhstahl?

In the meantime, I think getting the tus-middleware project to 1.0 would be best. For reference:

https://github.com/foobarable/tus-middleware/issues/1#issuecomment-194296099
https://github.com/tus/tus-android-client/issues/4#issuecomment-194293352

Acconut:
Actually, I believe it should already be possible to use tus-node-server with express. I haven't tested it thou:

```js
const server = new tus.Server();
server.datastore = new tus.FileStore({
    path: '/files'
});

app.all('/files/*', server.handle);
app.listen(PORT, HOST);
```
```

bhstahl:
Correct! The server is designed work as a standalone, or to be used with any node web framework (Express, [Restify](http://restify.com/#common-handlers-serveruse)) that uses the [`IncomingMessage`](https://nodejs.org/api/http.html#http_class_http_incomingmessage) class of the [`http`](https://nodejs.org/api/http.html#http_http) module.

@kvz unfortunately our Q1 upload team roadmap got a bit hijacked by other projects, but tus in production is still a Q1 item, which at this juncture might unfortunately flow into April. We haven't lost sight though, a tus node server running with a local & GCS FileStore is on the horizon!

kvz:
I stand corrected! And thanks for the update @bhstahl!

ReverseFlash28:
@Acconut its not working :(

bhstahl:
@ReverseFlash28 sorry, but that may be due to this still being a work in progress (which is why I haven't published it to NPM yet). You may want to check out https://github.com/foobarable/tus-middleware in the meantime, or contribute to the development here as well!

What version of Express are you running?

ReverseFlash28:
@bhstahl I already use that but it makes me error when my client in android

Acconut:
@bhstahl The reason why @ReverseFlash28 cannot use tus-middleware is that it does not support tus 1.0 which is a requirement for tus-android-client.

> its not working :(

What is not working? We need more information about what happened. Any errors?

ReverseFlash28:
@Acconut your code 

`const server = new tus.Server();
server.datastore = new tus.FileStore({
    path: '/files'
});

app.all('/files/*', server.handle);
[...]

PR #19, opened 2016-05-07
Document how to deploy the server as middleware

kvz:
See https://github.com/tus/tus-node-server/issues/16


bhstahl:
Great call! thanks 😄

PR #30, opened 2016-05-20
Custom file names

js3692:
User can now provide a hashing function to set a custom file name for each upload. When the hashing function is not provided, it defaults to a random `Uid`.

Example:
```javascript
const fileNameFromUrl = (req) => {
    return req.url.replace(/\//g, '-');
}

server.datastore = new tus.FileStore({
    path: '/files',
    namingFunction: fileNameFromUrl
});
```
Assuming the POST request was made to `/files`, the file name will now be `'-file'`


lib/handlers/HeadHandler.js
```diff
@@ -11,7 +11,7 @@ class HeadHandler extends BaseHandler {
      * @return {function}
      */
     send(req, res) {
-        const re = new RegExp('\\' + this.store.path + '\\/(\\w+)\/?'); // eslint-disable-line prefer-template
+        const re = new RegExp('\\' + this.store.path + '\\/([\\w\-]+)\/?'); // eslint-disable-line prefer-template
```
bhstahl: This was added just for that example hashing function from our convo on the issue, right? Maybe it should just match the non-whitespace (`\S`) rather than any alphanumeric character.

lib/stores/DataStore.js
```diff
@@ -37,13 +41,20 @@ class DataStore {
...
      */
-    create(file) {
-        if (!(file instanceof File)) {
-            throw new Error(`${file} is not a File`);
-        }
-        console.log(`[DataStore] create: ${file.id}`);
+    create(req) {
+        console.log('[DataStore] create');
+
+        return new Promise((resolve, reject) => {
```
bhstahl: Looking at this now, it probably makes sense for a `create` method to resolve with the model its creating. Do you think this should this resolve with the whole `File` rather than just the `file_name`?

lib/handlers/HeadHandler.js
js3692: Yes, that was to match the example function from the issue. I think `\S` would be more concise too :)

lib/stores/DataStore.js
js3692: Right, since the whole point of `create` is to create an instance of `File` with all the properties about its location, etc, it makes more sense to return the whole `File` rather than just the name. Edits coming soon!

lib/stores/DataStore.js
```diff
@@ -37,13 +42,26 @@ class DataStore {
...
+            const upload_length = req.headers['upload-length'];
+            const upload_defer_length = req.headers['upload-defer-length'];
+            const upload_metadata = req.headers['upload-metadata'];
+
+            if (upload_length === undefined && upload_defer_length === undefined) {
+                return reject(412);
```
bhstahl: These error code/messages should probably be constants, but I can do that in a separate PR :)

bhstahl:
Thanks @js3692 !!

avillarubia:
only renames to - hypen without extension. What I need is to save the file to its original name

PR #49, opened 2016-08-24
Remove HTTP as part of POST handler response

### LINKED ISSUE #48: HTTP assumed as part of POST response
Is there any particular reason why the POST handler includes 'http://' as part of it's response?
```javascript
const url = `http://${req.headers.host}${this.store.path}/${File.id}`;
```

Can this be changed to:
```javascript
const url = `//${req.headers.host}${this.store.path}/${File.id}`;
```

I'm running into a scenario where a POST is issued over HTTPS, and then the subsequent PATCH is over HTTP because of this.

stevewillard:
Removes `http://` prefix from POST handler response

Fixes #48

### DIFF
--- a/lib/handlers/PostHandler.js
+++ b/lib/handlers/PostHandler.js
@@ -15,7 +15,7 @@ class PostHandler extends BaseHandler {
     send(req, res) {
         return this.store.create(req)
             .then((File) => {
-                const url = `http://${req.headers.host}${this.store.path}/${File.id}`;
+                const url = `//${req.headers.host}${this.store.path}/${File.id}`;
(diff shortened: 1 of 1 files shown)


kvz:
I guess you could also omit protocol & hostname and the client will just re-use what it used in the first place?

bhstahl:
Yeah, agreed with @kvz .

@stevewillard if you can update and ensure the tests pass (not sure how [keyfile.json.enc](https://github.com/tus/tus-node-server/blob/master/keyfile.json.enc) was not included in that [last build](https://travis-ci.org/tus/tus-node-server/builds/154618988)), then I'll merge and bump!

PR #55, opened 2016-10-04
Add support for using tus on an express router

gerdus:
Basically changed the regex to match on req.originalUrl instead of req.url
and added req.baseUrl to creation Location to get correct URL.


bhstahl:
Hey @gerdus, thanks for submitting this PR. I'll work on why travis is not properly running the tests, but if you run this branch locally, you'll notice this introduces some breaking changes. If you fix them and are able to run `$ npm test` without any failures, I'll merge and bump.

gerdus:
Sorry about that. I fixed the code so the local tests pass.

PR #67, opened 2017-02-20
This GCS data store goes directly to the GCS resumable API, bypassing the google cloud SDK

m0rganic:


bhstahl:
Thanks for working on tus @m0rganic! I'm hesitant to replace the SDK that Google maintains to use GCS, since we'll have to keep up to date with future improvements ourselves. Have you considered submitting a PR to `google-cloud` then we can bump the version here when it gets merged?

m0rganic:
There shouldn't be much risk to that since we're going directly to the
google cloud JSON API and we can always have the GCSDataStore to fall back
on.

m0rganic:
That being said I'll check with the gcloud node maintainers and see if I can help out there. Although they may be resistant to artificially delaying the write operation to ensure filesize is reported properly - that seems much more like a tus-node-server requirement becasue its an artifact of how the client and server intereact when data is being chunked from client.

nfriedly:
Does this fix the 409 conflict errors that I'm getting after a chunk of data has been uploaded but google is still reporting a file size of 0?

m0rganic:
Yes I believe so

nfriedly:
Is there any way I can test this locally (aside from copying the changes over one-by-one)? I'd normally just npm install from your fork, but GitHub can't seem to find it:

screen shot 2017-05-25 at 10 07 43 am

m0rganic:
you should be able to roll with this pretty easily...

https://gist.github.com/m0rganic/dab293011ee328e45427cc0b06ca3076

bhstahl:
Thanks for adding this, but would #90 solve your problems @m0rganic ?

Issue #59, opened 2016-11-17
Consider server and browsers with creation url


bhstahl:
Right now the creation extension [omit's thee protocol](https://github.com/tus/tus-node-server/blob/be1e716f26945ef4bf8abb8d546a456fa356efa2/lib/handlers/PostHandler.js#L18) from a created url, which can cause some issues using tus-js-client server side (https://github.com/tus/tus-java-client/issues/10#issuecomment-258384118). We should consider a new approach that works in the browser and server

Acconut:
tusd will derive the protocol from the request (https://github.com/tus/tusd/blob/master/unrouted_handler.go#L633-L637). Later this value may be overwritten by the X-Forwarded-Proto header (https://github.com/tus/tusd/blob/master/unrouted_handler.go#L645-L647) which is commonly used by proxies. A similar approach may be useful tus-node-server.

bhstahl:
Like that option @Acconut, we'll follow suit

PR #73, opened 2017-09-20
Add protocol to location header in POSTs

erikhandersson:
The protocol in the location header was missing in POST
requests. The protocol is present in tusd.

### DIFF
--- a/lib/handlers/PostHandler.js
+++ b/lib/handlers/PostHandler.js
@@ -15,7 +15,7 @@ class PostHandler extends BaseHandler {
     send(req, res) {
         return this.store.create(req)
             .then((File) => {
-                const url = `//${req.headers.host}${req.baseUrl || ''}${this.store.path}/${File.id}`;
+                const url = `http://${req.headers.host}${req.baseUrl || ''}${this.store.path}/${File.id}`;
                 this.emit(EVENT_ENDPOINT_CREATED, { url });
                 return super.send(res, 201, { Location: url });
             })
--- a/test/Test-EndToEnd.js
+++ b/test/Test-EndToEnd.js
@@ -85,6 +85,7 @@ describe('EndToEnd', () => {
                 .expect(201)
                 .end((err, res) => {
                     assert.equal('location' in res.headers, true);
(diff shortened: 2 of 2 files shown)


kvz:
Good catch @erikhandersson! I wonder if we should loosen up the protocol instead tho in this regard. Interested to hear what @Acconut thinks of this. IIRC there was talk once about officially supporting completely relative paths (so `/files/xxx`) but perhaps we never poured that into protocol. Something slightly different than allowing protocol-relative paths (so `//example.com/files/xxx`) but perhaps worth discussing explicitly allowing both in the protocol as it seems it might solve some problems for people implementing it in the different infra scenarios out there

Acconut:
The protocol was originally not presented in the URL to let browsers make the decision whether to use HTTPS or HTTP, similar to how you can write `<script src="//myhost.com/script.js></script>` and the browser will pick the corresponding protocol. I believe this worked well for browsers but not other platforms (e.g. Android), so we decided to change this in #59 but it, apparently, hasn't been done yet.

Anyway, thank you for your PR but I believe this is not the optimal solution. You may want to check the X-Forwarded-Proto header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) first, to decide whether to use `http://` or `https://` in order to properly support proxies. tusd also supports the Forwarded header (https://github.com/tus/tusd/blob/master/unrouted_handler.go#L714-L750), which could also be helpful for you.

> I wonder if we should loosen up the protocol instead tho in this regard. Interested to hear what @Acconut thinks of this. IIRC there was talk once about officially supporting completely relative paths (so /files/xxx) but perhaps we never poured that into protocol. 

In fact, the protocol does allow absolute and relative URLs :) From the spec:

> The Server MUST set the Location header to the URL of the created resource. This URL MAY be absolute or relative.

It does not mention protocol-relative URLs but I have never seen them outside of tus-node-server for the Location header.

kvz:
> In fact, the protocol does allow absolute and relative URLs :) From the spec:

Ah okay, thanks for correcting me! 👌 All good then

erikhandersson:
I see. Thanks.

PR #78, opened 2017-11-30
patch url generation fix

vencax:
When I wanted to use express app for uploading:
```
const uploadApp = express();
uploadApp.all('*', server.handle.bind(server));
mainApp.use('/uploads', uploadApp);
```
it didn't work because generation of patch url was based on filestorage.path which IMHO should not have had nothing to do with url prefix of incomming request.

PS: usage of express would make much more easy even the rest of the app ;) What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,9 @@ server.datastore = new tus.FileStore({
 });
 
 var app = express();
-app.all('/files/*', function(req, res) {
-  server.handle(req, res);
-});
+const uploadApp = express();
+uploadApp.all('*', server.handle.bind(server));
+app.use('/uploads', uploadApp);
 app.listen(port, host);
 ```
 
--- a/demo/server.js
+++ b/demo/server.js
@@ -60,6 +60,18 @@ server.on(EVENTS.EVENT_UPLOAD_COMPLETE, (event) => {
     console.log(`[${new Date().toLocaleTimeString()}] [EVENT HOOK] Upload complete for file ${event.file.id}`);
 });
 
+// // this is the express stile ;)
+// const express = require('express');
+// const app = express();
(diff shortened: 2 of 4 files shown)


bhstahl:
Hey @vencax, thanks for offering up some help! I just pushed a change that _should_ get builds from forks to work (https://github.com/tus/tus-node-server/commit/650c86223a00987848e6f472519e830d3c59a5cc). Would you mind rebasing from upstream master so we can make sure the tests pass before proceeding?

Also, thanks for suggesting using express. I considered it when starting this project, but decided it would serve a larger community by remaining server framework agnostic.  I think we can easily make tus-node-server work with all node.js web frameworks, without imposing any specific ones on users. Happy to consider alternatives if the community disagrees though!

vencax:
OK, I have rebased the branch.
Concerning usage of express: it is defacto standard in node.js ecosystem and if used you don't have to reimplement thinks like e.g. BaseHandler.send or the regex in getFileIdFromRequest since express has already solve it in uniform way.

bhstahl:
Good stuff! Thanks for rebasing to ensure the tests pass, and for making this change. You can ignore the coverage decrease until I figure out a better way to run GCSDataStore tests from forks.

Agreed, however I personally don't feel it's enough code to impose a framework on the users of this server. But, if the community feels strong enough about adopting it I'm happy to consider.

bhstahl:
Released in [v0.2.6](https://github.com/tus/tus-node-server/releases/tag/v0.2.6). Thanks @vencax!

vencax:
Ok, I will send another PR with the changes I have mentioned earlier. Thanks for cooperation. Cheers

PR #79, opened 2017-11-30
Fix FileStore fire upload complete before file written to disk

gerdus:
FileStore currently has a bug where is fires the Upload Complete event before the filestream is finished. 
The consequence is that copying/moving the uploaded file in the upload complete event results in incomplete file. Happens especially on small files.


bhstahl:
Thanks @gerdus !

PR #96, opened 2018-03-10
Replace configstore with plain json files in upload path

gerdus:
configstore does a few things that are problematic:
- writes a file outside upload directory and with permissions so that group can't read/write
- file infos are never removed so it just grows larger

Moving to plain json file per upload in the upload directory enables you to clean the directory periodically of old files without disturbing in progress uploads. Same method that tusd uses.

### DIFF
--- a/lib/stores/FileStore.js
+++ b/lib/stores/FileStore.js
@@ -67,7 +67,7 @@ class FileStore extends DataStore {
             }
 
             const file = new File(file_id, upload_length, upload_defer_length, upload_metadata);
-            const filepath = this.binPath(file_id);
+            const filepath = this.filePath(file_id);
 
             return fs.open(filepath, 'w', (err, fd) => {
                 if (err) {
@@ -75,7 +75,14 @@ class FileStore extends DataStore {
                     return reject(err);
                 }
 
-                this.writeConfig(file);
+                try {
+                    this.writeFileInfo(file);
+                }
+                catch (infoWriteError) {
+                    log('[FileStore] create: Info Write Error', infoWriteError);
+                    return reject(infoWriteError);
+                }
+
 
                 return fs.close(fd, (exception) => {
                     if (exception) {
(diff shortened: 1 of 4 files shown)


lib/stores/FileStore.js
```diff
@@ -168,6 +167,32 @@ class FileStore extends DataStore {
...
+
+    getConfig(file_id) {
+        try {
+            const infoFilename = this.configPath(file_id);
+            const data = fs.readFileSync(infoFilename);
+            const file = JSON.parse(data);
+            return file;
+        }
+        catch (e) {
+            return null;
```
bhstahl: This should probably throw the error to in order to debug, and return an error code to the request?

lib/stores/FileStore.js
```diff
@@ -168,6 +167,32 @@ class FileStore extends DataStore {
             });
         });
     }
+
+    binPath(file_id) {
+        return `${this.directory}/${file_id}`;
+    }
+
+    configPath(file_id) {
```
bhstahl: I wonder if there is a better name than _config_, we don't need to beholden to it because the library was previous named prefixed with it.

lib/stores/FileStore.js
gerdus: the logic in write and getoffset expects config to be falsey if it doesn't exist.
write and getOffset currently do not handle the case when file info is missing/unreadable and caused alot of test failures.
I have added a rejection when creating the file since most failures would be there.
also now log the error.

lib/stores/FileStore.js
gerdus: in the last commit I renamed it getFileInfo/writeFileInfo; think it is more descriptive and is similiar to tusd GetInfo/writeInfo

mitjap:
@gerdus after some years your feature request is implemented. Or at least it is possible to use custom `configstore` for `FileStore`. 
See `MemoryConfigstore` as example how to implement custom `configstore`.
https://github.com/tus/tus-node-server/blob/f6670b04f661562f880f52a7e3e00c503cbe3d19/lib/configstores/MemoryConfigstore.js#L11
https://github.com/tus/tus-node-server/blob/f6670b04f661562f880f52a7e3e00c503cbe3d19/lib/stores/FileStore.js#L27

PR #97, opened 2018-03-27
more params added to EVENT_ENDPOINT_CREATED event

vencax:
Hi, this change allows more customisation and, to be honest, I need it :) I hope you'll consider it necessary as well.


Murderlon:
Thanks for the PR. This will be superseded by #336

Issue #98, opened 2018-03-29
Can't upload 0 Byte file


fentas:
From the tus protocol, tus should be able to upload 0 bytes files.

https://github.com/tus/tus-node-server/blob/master/lib/validators/RequestValidator.js#L21
`0` is included in the `non-negative integer` range.

I think it is not as easy as to change the `1` to a `0`?

---
As for the `tus-js-client` it should be able to upload/create 0 bytes file.
https://github.com/tus/tus-js-client/issues/106

PR #101, opened 2018-03-29
Enable 0 byte files

fentas:
fixes https://github.com/tus/tus-node-server/issues/98

---

According to @Acconut (https://github.com/tus/tus-js-client/issues/106#issuecomment-377223525), 0 bytes files should be handled directly in the `POST` (creation of a file) request.

In order to solve this, this pull request unblocks a file size of 0 bytes (`non-negative integer`).

The check if a file has 0 bytes will happen in every DataStores itself (within the `create` method).
I think updating current DataStores can be done in different pull requests.

### DIFF
--- a/lib/validators/RequestValidator.js
+++ b/lib/validators/RequestValidator.js
@@ -18,7 +18,7 @@ class RequestValidator {
 
     // The value MUST be a non-negative integer.
     static _invalidUploadLengthHeader(value) {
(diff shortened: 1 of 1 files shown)


bhstahl:
Released in [`v0.2.11`](https://github.com/tus/tus-node-server/releases/tag/v0.2.11), thanks @fentas!

Issue #108, opened 2018-05-17
Filenames with spaces and using namingFunction


PaddoSwam:
I tried to upload a file with spaces in the file name and it threw me an error, I debugged and found out that it didn't try to decode the file name.

Murderlon:
Hi, I updated the docs on the naming function, I think it's up to the implementer to create valid URL friendly names if they want custom names

PR #109, opened 2018-05-17
Filenames with spaces and using namingFunction

PaddoSwam:
Fixing #108 Untested in other scenario's so might need additional attention!


Acconut:
Firstly, thank you very much for your contribution. I still have a question about when this issue occurs. Is this caused because you return values which contain spaces form your namingFunction?

Murderlon:
See https://github.com/tus/tus-node-server/issues/108#issuecomment-1055110214

PR #120, opened 2018-07-17
Resolve dependency vulnerabilities in stringstream and base64url

spencern:
Resolves #119 
### Summary
This PR is a set of commits which remove or update dependencies in order to eliminate dependence on outdated versions of `stringstream@0.0.5` and `base64url@2.0.0` which both have reported vulnerabilities.

### Updated packages
- Updates outdated google-cloud/storage version from 1.1.1 to 1.7.0
- Updates request from ^2.8.5 to 2.8.7 and removes request from the list of direct dependencies as it's never used directly. 
Updating request to 2.87 removes the dependency on stringstream@0.0.5 which has a medium severity security vulnerability. https://snyk.io/vuln/npm:stringstream:20180511
- Removes unused package google-auto-auth@0.8.2. This package was listed as a direct dependency, but is not used in the package directly. This package is a dependency of `@google-cloud/storage` but since updating that package, we no longer need this version.
Removing this package removes 6 vulnerable paths to base64url@2.0.0 which is a High Severity vulnerability. https://snyk.io/vuln/npm:base64url:20180511
- Update jwa from 1.1.5 to 1.1.6 and ecdsa-sig-formatter from 1.0.9 to 1.0.10 
Removes dependency on base64url which has a high severity vulnerability reported. https://snyk.io/vuln/npm:base64url:20180511

Issue #119, opened 2018-07-17
Security vulnerabilities introduced through pinned version of @google-cloud/storage


spencern:
We use the `snyk` tool at ReactionCommerce to scan for vulnerabilities reported in npm packages.

We're seeing two different vulnerabilities introduced through `tus-node-server` and both of them are coming through the outdated dependency on @google-cloud/storage@1.1.1

A medium severity vulnerability on tunnel-agent: https://snyk.io/vuln/npm:tunnel-agent:20170305
```
tus-node-server@0.2.11 > @google-cloud/storage@1.1.1 > gcs-resumable-upload@0.7.7 > google-auto-auth@0.6.1 > gcp-metadata@0.1.0 > retry-request@1.3.2 > request@2.76.0 > hawk@3.1.3 > hoek@2.16.3
```

A low severity vulnerability on hoek: https://snyk.io/vuln/npm:hoek:20180212
```
tus-node-server@0.2.11 > @google-cloud/storage@1.1.1 > gcs-resumable-upload@0.7.7 > google-auto-auth@0.6.1 > gcp-metadata@0.1.0 > retry-request@1.3.2 > request@2.76.0 > tunnel-agent@0.4.3
```

After splunking through the @google-cloud/storage releases, it seems that if you can update to v1.7.0 of `@google-cloud/storage` that should bump the version of `request` enough to eliminate these vulnerabilities. 

Version 0.11 of `gcs-resumable-upload` contains a different auth library than google-auto-auth which should also resolve the vulnerabilities, but is not yet available in any `@google-cloud/storage` releases yet.

Acconut:
Thank you very much reporting this. Would you mind opening a pull request for upgrading `@google-cloud/storage`? This patch can then be included in the 0.3.0 release which is right around the corner.

spencern:
@Acconut There's a new vulnerability reported in the `cryptiles` library that's included in the `@google-cloud/storage@1.1.1` release. That vuln is also resolved by #120. 

When do you think you'll be able to cut a release?

Acconut:
0.3.0 has just been released right now.

spencern:
Thank you!

PR #121, opened 2018-07-17
Spencer remove object assign dep

spencern:
**Note**: This is an unsolicited change that I noticed opportunity for when working on resolving dependency vulnerabilities for #119 - No trouble at all if you don't accept this PR.

### Summary
Refactor the GCSDataStore to use `Object.assign()` instead of the npm package `object-assign` polyfill. Removes direct dependency on `object-assign`

### Rationale
As this same file is using `let` and arrow functions, both of which require node >=4, it seems like a safe change to make.

This PR is built on top of #120 - this is a small change otherwise.


Acconut:
The less dependencies the better :) Thank you very much

Issue #100, opened 2018-03-29
Location should send relative url


kmjayadeep:
In POST request, the response header should contain Location as relative url. In my case I was running application behind nginx proxy. So tus was returning 127.0.0.1/file_url in the Location. If it was relative url, it should be easy to handle for the client

Acconut:
Historically, not all clients (especially on Android and iOS) have dealt very well if the URL returned in the Location header is not absolute. This is why we usually prefer to provide no relative URL. However, we would be happy to see a PR to change this.

gormed:
@bhstahl I think this can be closed now :)

bhstahl:
thanks @gormed  :)

PR #128, opened 2018-08-16
Relative location support

gormed:
Added relative location support like #100 suggests, i hope it is okay :)


bhstahl:
Thanks! Do you mind adding a test demonstrating the expected behavior?

gormed:
Will do, do you have a specific place for options to test, or should i just add a new test-file?

gormed:
Alright, i moved the option to the DataStore and also tested it there, since it is a general setting. I also added a E2E test for the FileStore with relativeLocation option. Should i add one for the GCSDataStore too?

test/Test-EndToEnd.js
```diff
@@ -216,6 +216,184 @@ describe('EndToEnd', () => {
...
+            server = new Server();
+            server.datastore = new FileStore({
+                path: STORE_PATH,
+                // configure the store to return relative path in Location Header
+                relativeLocation: true
+            });
+            agent = request.agent(server.listen());
+        });
+
+        after((done) => {
```
bhstahl: This is unnecessary because it is already [declared for every test](https://github.com/tus/tus-node-server/blob/76ac93544a8c4fc0878dc00cfdd8c07f8dd09f49/test/Test-EndToEnd.js#L56-L67)

test/Test-EndToEnd.js
```diff
@@ -216,6 +216,184 @@ describe('EndToEnd', () => {
...
+                }
+
+                // clear the config
+                server.datastore.configstore.clear();
+                return done();
+            });
+        });
+
+        describe('HEAD', () => {
+            it('should 404 file ids that dont exist', (done) => {
```
bhstahl: No need to re-test this, just the new functionality

test/Test-EndToEnd.js
```diff
@@ -216,6 +216,184 @@ describe('EndToEnd', () => {
...
+                .end((err, res) => {
+                    assert.equal('location' in res.headers, true);
+                    assert.equal(res.headers['tus-resumable'], TUS_RESUMABLE);
+                    // Save the id for subsequent tests
+                    file_to_delete = res.headers.location.split('/').pop();
+                    done();
+                });
+            });
+
+            it('should create a file and respond with its _relative_ location', (done) => {
```
bhstahl: this is really the only new test you need, regressions would be caught in the existing tests

gormed: ok, i removed the unnecessary tests

gormed:
@bhstahl Can you have a look into it, and maybe bump a new version to npm? :pray:

test/Test-EndToEnd.js
```diff
@@ -216,6 +216,41 @@ describe('EndToEnd', () => {
         });
     });
 
+    describe('FileStore with relativeLocation', () => {
+        let file_id;
+        let deferred_file_id;
```
bhstahl: `deferred_file_id` is just for the End To End test on deferring the file location. Do you need to store `file_id` and  `deferred_file_id` in a variable?

gormed: Ah, no just remains from the whole test i copied before, i'll remove it.

Acconut:
@bhstahl @gormed Is this PR ready?

gormed:
I fixed the tests, i think it is ready, let @bhstahl decide if there needs to be done sth more.

bhstahl:
`v0.3.2` released 🎉

PR #89, opened 2018-01-08
Initial Amazon S3 support

radev:

PR #130, opened 2018-08-23
GSC Datastore: Add support for contentType property

### LINKED ISSUE #57: GSC Datastore does not support "contentType" property
There should be a way to set contentType on the uploaded object as such...

  var stream = file.createWriteStream({
    metadata: {
      contentType: <mimetype>
    }
  });

tkrugg:
closes #57

This PR follows up on #57 

image

I couldn't get test/Test-GCSDataStore.js so i'll trust the travis output to fix the tests

### DIFF
--- a/lib/stores/DataStore.js
+++ b/lib/stores/DataStore.js
@@ -81,7 +81,7 @@ class DataStore extends EventEmitter {
      */
     write(req) {
         log('[DataStore] write');
-        return new Promise((resolve, reject) => {
+        return new Promise((resolve) => {
             // Stub resolve for tests
             const offset = 0;
 
@@ -92,7 +92,7 @@ class DataStore extends EventEmitter {
 
     /**
      * Called in HEAD requests. This method should return the bytes
-     * writen to the DataStore, for the client to know where to resume
+     * written to the DataStore, for the client to know where to resume
      * the upload.
      *
      * @param  {string} id     filename
@@ -115,6 +115,9 @@ class DataStore extends EventEmitter {
      * @return {Object}                 metadata as key-value pair
      */
     _parseMetadataString(metadata_string) {
+        if (!metadata_string || !metadata_string.length) {
+            return {};
+        }
         const kv_pair_list = metadata_string.split(',');
 
         return kv_pair_list.reduce((metadata, kv_pair) => {
--- a/lib/stores/DataStore.js
+++ b/lib/stores/DataStore.js
@@ -107,6 +107,27 @@ class DataStore extends EventEmitter {
             return resolve({ size: 0, upload_length: 1 });
         });
     }
+
+    /**
+     * Parses the Base64 encoded metadata received from the client.
+     *
+     * @param  {String} metadata_string tus' standard upload metadata
+     * @return {Object}                 metadata as key-value pair
+     */
+    _parseMetadataString(metadata_string) {
+        const kv_pair_list = metadata_string.split(',');
+
+        return kv_pair_list.reduce((metadata, kv_pair) => {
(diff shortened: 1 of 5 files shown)


lib/stores/GCSDataStore.js
```diff
@@ -92,7 +92,9 @@ class GCSDataStore extends DataStore {
 
             const file = new File(file_id, upload_length, upload_defer_length, upload_metadata);
             const gcs_file = this.bucket.file(file.id);
+            const parsedMetadata = this._parseMetadataString(upload_metadata);
             const options = {
+                contentType: parsedMetadata.type.decoded,
```
Acconut: tus-js-client uses the `filetype` metadata key to specify the file's type (see https://github.com/tus/tus-js-client#example). I would like it if we only had one property to do so. Would you mind changing it?
Besides that, a check to see if `parsedMetadata.type` actually exists might be necessary.

lib/stores/GCSDataStore.js
```diff
@@ -131,9 +133,11 @@ class GCSDataStore extends DataStore {
             return new Promise((resolve, reject) => {
                 const file = this.bucket.file(file_id);
 
+                const parsedMetadata = this._parseMetadataString(data.upload_metadata);
                 const options = {
                     offset,
                     metadata: {
+                        contentType: parsedMetadata.type.decoded,
```
Acconut: Again, a check to see if `parsedMetadata.type` actually exists might be necessary.

tkrugg: you're right 👍 will add it

Acconut:
> I was going to write it but I couldn't get the GCS test suite to run locally. It does run on travis. I may be missing some env variables. Are there a few steps to follow before running them locally?

Yes, there are a few steps to follow. They are not directly documented anywhere but that's on my list to improve. The basic steps should be:
- Create a bucket on GCS
- Save the keyfile from Google Cloud as `keyfile.json` in the root of the tus-node-server directory
- Adjust the project Id and bucket name values in `test-GCSDataStore.js`: https://github.com/tus/tus-node-server/blob/9ac6bf9e4964f86b93a36f765405ffc43f535c5e/test/Test-GCSDataStore.js#L17-L19
- Upload the file `test/test.mp4` as `dont_delete_this_file.mp4` to your bucket

I think that should be all the steps. Let me know if they work for you.

> Thank you very much for taking the time to review this PR. I appreciate.

You're welcome. Thank you very much for contribution.

lib/stores/GCSDataStore.js
tkrugg: @Acconut small reminder about my question:
>  was going to write it but I couldn't get the GCS test suite to run locally. It does run on travis. I may be missing some env variables. Are there a few steps to follow before running them locally?

Acconut: I responded to it in the main thread two days ago: https://github.com/tus/tus-node-server/pull/130#issuecomment-422711001

lib/stores/DataStore.js
```diff
@@ -106,6 +106,27 @@ class DataStore extends EventEmitter {
             return resolve({ size: 0, upload_length: 1 });
         });
     }
+
```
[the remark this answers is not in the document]
tkrugg: referencing related issue: https://github.com/tus/tus-node-server/pull/89/files#diff-cf34f7e211aecb1cd3ae5f05a5cde2d8R242

Acconut:
@tkrugg Were you able to have a look at the GCS-tests yet? Is this PR ready from your side?

Murderlon:
Closing this as stale and the source has now significantly diverged. Thanks for taking the time to contribute.

PR #132, opened 2018-09-13
fix relative location url

sh:
hey, thanks for #128 

with `relativeLocation: true` the returned location is only `{uuid}`
it should return the same url just without scheme + host

with patch applied:
`relativeLocation: false` => `//localhost/tus/files/{uuid}`
`relativeLocation: true` => `/tus/files/{uuid}`

### DIFF
--- a/lib/handlers/PostHandler.js
+++ b/lib/handlers/PostHandler.js
@@ -16,7 +16,7 @@ class PostHandler extends BaseHandler {
     send(req, res) {
         return this.store.create(req)
             .then((File) => {
-                const url = this.store.relativeLocation ? `${File.id}` : `//${req.headers.host}${req.baseUrl || ''}${this.store.path}/${File.id}`;
+                const url = this.store.relativeLocation ? `${req.baseUrl || ''}${this.store.path}/${File.id}` : `//${req.headers.host}${req.baseUrl || ''}${this.store.path}/${File.id}`;
 
                 this.emit(EVENT_ENDPOINT_CREATED, { url });
                 return super.send(res, 201, { Location: url });
--- a/test/Test-EndToEnd.js
+++ b/test/Test-EndToEnd.js
@@ -256,7 +256,7 @@ describe('EndToEnd', () => {
                     // the location header is not absolute
                     assert.equal(res.headers.location.indexOf("//") === -1, true);
                     // and contains the store path
-                    assert.equal(res.headers.location.indexOf(STORE_PATH) > -1, false);
+                    assert.equal(res.headers.location.indexOf(STORE_PATH) > -1, true);
                     done();
                 });
             });
--- a/test/Test-EndToEnd.js
+++ b/test/Test-EndToEnd.js
@@ -253,7 +253,9 @@ describe('EndToEnd', () => {
                 .end((err, res) => {
                     assert.equal('location' in res.headers, true);
                     assert.equal(res.headers['tus-resumable'], TUS_RESUMABLE);
-                    // the location header is relative to the store path
+                    // the location header is not absolute
+                    assert.equal(res.headers.location.indexOf("//") === -1, true);
+                    // and contains the store path
                     assert.equal(res.headers.location.indexOf(STORE_PATH) > -1, false);
                     done();
                 });


sh:
friendly ping @bhstahl @gormed, would love to see this fix released :)

Acconut:
Thank you very much for this contribution. I agree with you that it is unintuitive when `relativeLocation` currently returns. It depends on if "relative" is interpreted as relative to the entire path (which is currently the case) or just relative to the hostname and protocol (as you propose it here). However, I am unsure if that may be the intention behind #128, so a comment from @gormed would be nice.

gormed:
Have thought the same. the question was, if it is relative to the store path or to the base url. 

The spec is not clear here: https://tus.io/protocols/resumable-upload.html

> The Server MUST acknowledge a successful upload creation with the 201 Created status. The Server  MUST set the Location header to the URL of the created resource. This URL MAY be absolute or relative.

The clients know the store path, and the tus-js-client handles this without any adjustments, so it was fine for me. But i know that the go client does not behave like that. In my opinion, the clients know the endpoint, and can build the correct url by themselfs, so they should be only the file hash (@sh why uuid?). But the spec should more specific here, and then we can adjust to it.

sh:
Thanks for having a look!

I'm happily using [@uppy/tus](https://uppy.io/docs/tus/) for my uploads, which uses [tus-js-client](https://github.com/tus/tus-js-client) internally as well.

> The Server MUST acknowledge a successful upload creation with the 201 Created status. The Server  MUST set the Location header to the URL of the created resource. This URL MAY be absolute or relative.

A Location header should contain a valid path to an existing resource.

A few example requests with the current master branch:

Absolute (`relativeLocation: false`)
```
# get upload url
POST http://localhost/files
Status: 201
Location: //localhost/files/filename

Resolved: http://localhost/files/filename (OK)

# start upload
PATCH: http://localhost/files/filename
Status: 204
```


Relative (`relativeLocation: true`):
```
# get upload url
POST http://localhost/files
Status: 201
Location: filename

# see below why
Resolved: http://localhost/filename (!)

# start upload
PATCH: http://localhost/filename
Status: 404
```

**According to the specs this location resolution behavior is correct and that's where it breaks for me.**


From [7.1.2 Location - RFC7231](https://tools.ietf.org/html/rfc7231#section-7.1.2)

> The field value consists of a single URI-reference.  When it has the
> form of a relative reference ([[RFC3986], Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2)), the final
> value is computed by resolving it against the effective request URI
> ([[RFC3986], Section 5](https://tools.ietf.org/html/rfc3986#section-5)).


Example based on [5.4 Reference Resolution Examples - RFC3986](https://tools.ietf.org/html/rfc3986#section-5.4)

```
POST http://a/b/c/d;p?q
Location: g

Resolved: "http://a/b/c/g/" (!)
```


---

By changing the current implementation of relativeLocation it will resolve to a working url.

```
# get upload url
POST http://localhost/files
Location: /files/filename

Resolved: http://localhost/files/filename (OK)

# start upload
PATCH: http://localhost/files/filename
Status: 204
```

I'm not sure about the wording `relativeLocation` for this though :)

Besides my current setup consists of loadbalancer => tus-node-server, I could only get it to work with the changes made in this PR.

gormed:
Ah yes, i had the same issue and fixed it with the trailing slash. 

> Personally, when I think about an absolute URL in the web context, it should contain a protocol, hostname/IP and the entire path. A relative URL on the other hand does not need to have a protocol or hostname but should have the entire path. Please be aware that this is just my view to how I am used to it and you can also rationally in another way.

/agree - i was not sure in the PR and thought that the trailing slash worked fine (for me :sweat_smile:)

Let me test that with our implementation, and who should review it, do i have the rights for that?

bhstahl:
@gormed with the invite just sent to your inbox, you will have access!

sh:
I've rebased my patch against the current master.

Would love to see a merge + release at some point, it's a pita to debug locally without this fix :)

Acconut:
Apologies for the inactivity, I was waiting for a review from another person. Anyways, I just merged this PR, so hopefully debugging will be easier for you now.

PR #136, opened 2018-11-05
Take chunk/part to upload to S3 from req.body or req

saschbro:
`S3Store` expects the incoming request (`req`) to be a readable stream and forwards it to AWS SDK's `uploadPart` function. In case stream is already consumed and not readable anymore upload to S3 fails. In my use case, I check uploaded content in Express middleware and parse uploaded file data to `req.body` with `bodyParser` before request is forwarded to `tus-node-server`. In order to work I extended `_processUpload` to handle chunk/part data in `req.body` and fall back to `req` (stream) if not set.

### DIFF
--- a/lib/stores/S3Store.js
+++ b/lib/stores/S3Store.js
@@ -257,19 +257,19 @@ class S3Store extends DataStore {
     /**
      * Uploads a part/chunk to S3 from a temporary part file.
      *
-     * @param  {Object}          metadata            upload metadata
-     * @param  {Stream}          read_stream         incoming request read stream
-     * @param  {Number}          current_part_number number of the current part/chunk
-     * @return {Promise<String>}                     which resolves with the parts' etag
+     * @param  {Object}               metadata            upload metadata
+     * @param  {Buffer|Stream|String} data                upload part/chunk data
+     * @param  {Number}               current_part_number number of the current part/chunk
+     * @return {Promise<String>}                          which resolves with the parts' etag
      */
-    _uploadPart(metadata, read_stream, current_part_number) {
+    _uploadPart(metadata, data, current_part_number) {
         return this.client
             .uploadPart({
                 Bucket: this.bucket_name,
                 Key: metadata.file.id,
                 UploadId: metadata.upload_id,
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for this PR. Would you mind explaining your need to parse the uploaded file data before passing it to tus-node-server? I have never seen this approach before and wanted to understand you thoughts.

Besides that I have to say that I don't have enough knowledge in the S3Store to judge if your changes are safe, maybe @bhstahl or @rictorres can help.

PS: The linter complained about something: https://travis-ci.org/tus/tus-node-server/builds/451035436?utm_source=github_status&utm_medium=notification

saschbro:
The reason we want to parse an uploaded file before passing it to tus-node-server is to check the file's media type based on its content. That way we can directly respond with error code to one of the PATCH requests the file parts are uploaded with. At the moment, we have to listen to tus-node-server's UPLOAD_COMPLETE event, download the file from S3 again and check it. In case media type isn't a valid one, we cannot directly transfer this information back to the client from within the listener (however, this shouldn't happen by "normal" use of our client).

Acconut:
That's an interesting use case, thanks for explaining it. Did you account for situations when the PATCH request body does not start at the file's beginning? This can happen when the upload is interrupted and the upload is resumed for the last known position. In this case you cannot check the media type since there is not file signature.

Anyways, that's just the first thought I had. As I said before I don't have the personal knowledge about the S3Store to approve this PR but I plan on investing more time in the coming days into S3Store and will hopefully come back with more results to you.

rictorres:
managed to have a proper look at this pr.

@saschbro i totally understand your use case although i would suggest configuring this behavior when initializing the s3store. just to make it clear for our users and also to avoid _magic_ occurring internally ☺️ 	

what do you think?

also, are you able to partially consume the initial stream, store it in memory, validate the media type, pass stored/incoming data to a stream and then send it to tus? it does sound a bit complicated, just wanted to analyze all our options :)

thanks!

Murderlon:
Talked with @Acconut and we feel this might be too much of an edge case to merge this in.

PR #146, opened 2019-02-28
Updated dependencies to recent versions, fixed tests on Windows

DvdGiessen:
As I'll be looking into using `tus-node-server` in a Node.js based project the upcoming days I just spend some time updating the dependencies, checking which changes these included and making sure it doesn't impact any of the functionality. Tests have been updated and run flawlessly again, now including on Windows (and of course also still on Linux).


DvdGiessen:
In my original comment I noted that we might have to update the minimum required version, assuming Node 4 support might still be a thing. Turns out Node 6 actually is actually the defined minimum, which works fine even with the updated dependencies. :)

Pushed to fix a broken build in Travis, and then pushed again since didn't properly sign my commits after the rebase. Should be all green now. :)

DvdGiessen:
Updated the included demo to use the latest demo code from `tus-js-client`.

Acconut:
Thank you a lot for your efforts, it's greatly appreciated! We currently don't have many resources for tus-node-server available, so PRs like this one helps us very much.

PR #157, opened 2019-11-08
FIX: PATCH receives 404 with relativeLocation set to true

patrykcieszkowski:
Issue: PATCH requests receives `404 -  The file for this url was not found` error.

The `getFileIdFromRequest` method doesn't consider `relativeLocation` value, and always tries to compare the request URL with file's full path.


patrykcieszkowski:
Seems like I've misunderstood the purpose of #128.

PR #162, opened 2020-01-14
add typescript

kyoungduck:
I add typescript module "index.d.ts" to be compatible with TS.
It's not only can use JS but also can use TS.

I already applied it to my project.
please watch my PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,149 @@
+import { EventEmitter } from "events";
+import http from 'http';
+
+/**
+ * arguments of constructor which in class extend DataStore
+ */
+declare interface DataStoreOptions{
+    path: string;
+    namingFunction?: (req : http.IncomingMessage) => string;
+    relativeLocation?: string;
+}
+
+declare interface FileStoreOptions extends DataStoreOptions{
+    directory?: string;
+}
+
+declare interface GCStoreOptions extends DataStoreOptions{
+    bucket: string;
+    projectId: string;
+    keyFilename: string;
+}
+
+declare interface S3StoreOptions extends DataStoreOptions{
+    accessKeyId: string;
+    secretAccessKey: string;
+    bucket: string;
+    tmpDirPrefix: string;
+    partSize: number;
+}
+
+declare class File{
+    id: string;
+    upload_length: any;
+    upload_defer_length: any;
+    upload_metadata: any;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+    write(req: http.IncomingMessage, file_id?: string, offset?: number): Promise<any>;
+    getOffset(id: string): Promise<any>;
+}
+
+/**
+ * file store in local storage
+ */
+export declare class FileStore extends DataStore{
+    constructor(options : FileStoreOptions);
+    _checkOrCreateDirectory(): any;
```
Acconut: Why are the private methods (prefixed with an underscore) included in here? Is this necessary?

index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+    set datastore(store: DataStore);
+    get(path: string, callback: Function): any;
+    handle(req: http.IncomingMessage, res: http.ServerResponse): http.ServerResponse;
+    listen(): http.Server;
+}
+
+export declare const ERRORS: {
+    MISSING_OFFSET: {
+        status_code: 403,
+        body: 'Upload-Offset header required\n',
```
Acconut: Why are the constant values in here? If I understand it correctly, those values are duplicated from the actual JS code.

kyoungduck: I'm updating code now.
And I have a question somthings.

I understand that developers use constants related to events.
So I need to add constant about event at ts module.
Because when developer use ts module, ts compiler only find the constant at ts module.
TS compiler doesn't find js module.(I'm talking about when developer use "import" syntax)

My question is does developer who using our package use ERRORS constant?
If it's not, I'll remove the ERROR constant.

Acconut: I understand that you must declare the constants in the .d.ts-file, however, I think it should be possible to keep the constant's value outside of .d.ts. So that the constant's value is still only present in the .js file and the .d.ts file just says "hey, the ERRORS constant has this data type".

This is not only about ERRORS but also other constants like EVENT_ENDPOINT_CREATED which have their value inside .d.ts

index.d.ts
kyoungduck: I remove all method which starts with _

index.d.ts
kyoungduck: I remove constant's value and leave only declare.

Acconut:
Thank for very much!

PR #183, opened 2021-05-27
resumable uploads with GCS 

mitjap:
My solution is to write to temporary file on subsequent writes and combine with original file. Temporary file is discarded immediately after write.  

I added test for this behavior and improved some other test. Many tests did not remove files after.

Resolves https://github.com/tus/tus-node-server/issues/37

PR #185, opened 2021-05-28
implementation of creation-with-upload extention

mitjap:
When `Content-Type: application/offset+octet-stream` header is present also start upload. I did not want to start file upload when there is no `Content-Type` header as is implemented in goloang, beause that would make additional requests to store (GCS, S3, ...) also in requests when there is no data to upload.


Acconut:
Thank you very much for this PR! It is a bit outside of my comfort zone as I am not familiar with this code base. However, how would you feel about merging this on your own? See https://github.com/tus/tus-node-server/pull/183#issuecomment-856313062 :)

Issue #68, opened 2017-03-13
Events & Errors should be implemented in the handler instead of DataStore's


bhstahl:
Currently, events like the `EVENT_FILE_CREATED` are emitted from the DataStore level ([like so](https://github.com/tus/tus-node-server/blob/7f7b9b8dc5aaddc672b1488199023e8a65b4de0f/lib/stores/DataStore.js#L65)), and Error's that include response information are thrown from the DataStore level as well ([like so](https://github.com/tus/tus-node-server/blob/7f7b9b8dc5aaddc672b1488199023e8a65b4de0f/lib/stores/DataStore.js#L59)).

Since neither of these are the concern of the DataStore, and rather the HTTP request handler, they should be abstracted one level above to the handler. The DataStore should be purely focused on reading & writing.

This will also make #61 possible.

ThomasG77:
I've seen your last commit https://github.com/tus/tus-node-server/commit/faf40f0cc87bffc3b4ffe1632d2aadcee9537f26 and I'm wondering why we can't access to the `req` and `resp` when event like `EVENT_FILE_CREATED` are fired.

My use case explaining my suggestion:

I'm using the server with Express. I have an authentication with a JWT token where I have a user_id.
For this purpose, I have a header `authorization` with a value `bearer my_jwt_token`.
In this context, I need access to `req.headers` to be able to say to the system, for this created file, assign the original name, an id and a size with the authenticated user deduced from the token.
I need to set a returned content. So, I also need access to the response.

I don't say change had to be made as my use case may not fit most people usages. It can at least help you understanding why I think exposing request and response in events should be done IMO.

PR #186, opened 2021-05-28
Redesign stores for better separation of concerns

mitjap:
Datastores's responsibility should not be handling of http requests but working with readable streams and File class. This PR is just a first step towards this goal, it is probably not working and tests fail miserably. I'd like to get your opinion about these changes before I spend too much time.

Related to #68.


Acconut:
This looks very interesting, thank you! However, would you mind explaining what the actual changes are? Then I can try to provide some valuable feedback :)

Acconut:
I agree, those are very sensible changes and mimic a similar data storage interface as present in tusd! They will definitely make it easier to develop new tus features independent of the storage layer. Very forward thinking of you, thank you very much!

mitjap:
@Murderlon I think this PR is still worth working on. I can do some more work but my main issue is that this change breaks almost all existing tests. Is this something you could help me with?

Murderlon:
> @Murderlon I think this PR is still worth working on. I can do some more work but my main issue is that this change breaks almost all existing tests. Is this something you could help me with?

Currently the tests seem to fail and something unrelated, it can't decrypt the secret. I'll look into that. If you want to test locally, you will need credentials for S3 and GCS. If that's not possible for you, I can help with the tests at a later stage. I do agree with the proposed changes and discussions from this PR 👍

Murderlon:
@mitjap I fixed the tests in #267. If you merge and resolve conflicts we can start to take a look at the tests.

mitjap:
I've made rough changes so that you can have better understanding of what changes I propose. Main goal is still to simplify stores. Store's only responsibility now is to implement `create`, `write`, `delete` and `getOffset` functions and nothing more. Ideally I would also separate configstore for metadata but this can be left for next PR :) After this PR is finished stores can be easily extracted to separate packages/repositories.

lib/Server.js
```diff
@@ -9,24 +9,40 @@
...
+            throw new Error('\'path\' is not defined must have a path');
+        }
+        if (options.namingFunction && typeof options.namingFunction !== 'function') {
+            throw new Error('\'namingFunction\' must be a function');
+        }
+
+        this.options = { ...options };
+        if (!this.options.namingFunction) {
+            this.options.namingFunction = Uid.rand;
+        }
```
mitjap, line 43: Move this to PostHandler as it is only used there. Without `creation` extension it is not even needed.

lib/handlers/DeleteHandler.js
```diff
@@ -13,13 +13,13 @@ class DeleteHandler extends BaseHandler {
      */
     send(req, res) {
         const file_id = this.getFileIdFromRequest(req);
```
mitjap, line 15: this should return rejected promise not throw synchronously

Murderlon:
Can you mark as ready for review when it's ready? I'm also surprised to see this PR still doesn't have access to GH secrets after my PR 🤔

lib/handlers/HeadHandler.js
```diff
@@ -12,49 +12,50 @@ class HeadHandler extends BaseHandler {
...
+            }
+
+            return res.end();
+        }
+        catch (error) {
+            log('[HeadHandler]', error);
+            const status_code = error.status_code || ERRORS.UNKNOWN_ERROR.status_code;
+            const body = error.body || `${ERRORS.UNKNOWN_ERROR.body}${error.message || ''}\n`;
+            return super.send(res, status_code, {}, body);
         }
```
mitjap, line 58: This  error handling code is identical in all handler and could be moved to `Server::handle(req, res)` function.

lib/stores/S3Store.js
```diff
@@ -4,6 +4,7 @@ const assert = require('assert');
 const os = require('os');
 const DataStore = require('./DataStore');
 const { FileStreamSplitter } = require('../models/StreamSplitter');
+const Metadata = require('../models/Metadata');
```
mitjap, line 7: this can be removed, it is not used

lib/stores/S3Store.js
```diff
@@ -435,7 +447,10 @@ class S3Store extends DataStore {
 
     create(file) {
         return this._bucketExists()
-            .then(() => this._initMultipartUpload(file))
+            .then(() => {
+                return this._initMultipartUpload(file);
+            })
```
mitjap, line 452: this change is not needed, next line is though

Murderlon:
If we merge now then the tests finally run properly from forks (tested in your other PR)

lib/stores/S3Store.js
```diff
@@ -147,7 +154,9 @@ class S3Store extends DataStore {
 
                 return data.UploadId;
             })
-            .then((upload_id) => this._saveMetadata(file, upload_id))
+            .then((upload_id) => {
+                return this._saveMetadata(file, upload_id);
+            })
```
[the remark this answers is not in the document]
Murderlon: Why not?

test/Test-EndToEnd.js
```diff
@@ -13,9 +13,9 @@ const { Storage } = require('@google-cloud/storage');
 const TUS_RESUMABLE = require('../lib/constants').TUS_RESUMABLE;
 
 const STORE_PATH = '/test/output';
-const PROJECT_ID = 'tus-node-server';
-const KEYFILE = path.resolve(__dirname, '../keyfile.json');
-const BUCKET = 'tus-node-server-ci';
+const PROJECT_ID = 'supervisor-3dsurvey';
+const KEYFILE = '/home/mitjap/Projects/rune-panorama/supervisor-3dsurvey-editor-b9c28cd4108c.json';
+const BUCKET = 'mitjap-test-tus-node-server';
```
Murderlon, line 18: We should change this back. Ideally these values come from env as well so we can use `npm run test:local` locally and CI will inject the vars so nothing has to change in the codebase

test/Test-PostHandler.js
```diff
@@ -1,61 +1,177 @@
...
+            });
+
+            it('must 412 if the Upload-Length and Upload-Defer-Length headers are both present', async() => {
+                const handler = new PostHandler(fake_store, { namingFunction: () => '1234' });
+                req.headers = { 'upload-length': '512', 'upload-defer-length': '1'};
+                await handler.send(req, res);
+                assert.equal(res.statusCode, 412);
+            });
+
+            // it('should reject if both upload-length and upload-defer-length are not provided', async function () {
```
Murderlon, line 76: this can be removed

test/Test-Stores.shared.js
```diff
@@ -1,192 +1,137 @@
...
-      const namingFunction = (incomingReq) => incomingReq.doesnotexist.replace(/\//g, '-')
-      this.server.datastore.generateFileName = namingFunction
-      return assert.rejects(() => this.server.datastore.create(req), { status_code: 500 })
-    })
-
-    it('should create a file, process it, and send file created event', async function () {
-      const fileCreatedEvent = sinon.fake()
-
-      this.server.on(EVENTS.EVENT_FILE_CREATED, fileCreatedEvent)
+    const file = new File('1234', '1000', undefined, 'filename d29ybGRfZG9taW5hdGlvbl9wbGFuLnBkZg==,is_confidential');
```
Murderlon, line 27: this makes more sense indeed than `req` object's

test/Test-Stores.shared.js
```diff
@@ -1,192 +1,137 @@
...
+
+    it('should write a stream and resolve the new offset', async function () {
+      await this.datastore.create(file);
+
+      const readable = fs.createReadStream(this.testFilePath);
+      const offset = await this.datastore.write(readable, file.id, 0);
+      assert.equal(offset, this.testFileSize);
+    });
+
+    // TODO: Does this test make sense? Should the store just report an error, and force the server to get offset via `getOffset`?
```
Murderlon, line 106: It feels a bit off indeed when an underlying error occurred we still resolve an offset as if everything is fine. But not sure if `getOffset` still works after an error, depends on clean up I suppose

Murderlon (commented):
Massive PR, thanks! Not that many comments as I mostly see repetition of moving the same things around, which LGTM. Improvements to tests make sense too. I agree with some comments you reviewed yourself of things that can still happen.

test/Test-PatchHandler.js
```diff
@@ -105,6 +105,42 @@ describe('PatchHandler', () => {
                     assert.equal(res.statusCode, 204);
                 });
         });
+
+        // it('should open a stream, resolve the new offset, and emit upload complete', function (done) {
```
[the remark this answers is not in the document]
mitjap: This was moved from shared tests. I think we have separate tests elsewhere (Test-Server.js) that cover this test's scenario.

test/Test-Stores.shared.js
mitjap: As per specification server should try to save as mush as it can. If there is an error some data might be lost and probably an error reported. HEAD request should always succeed IMO and allow client to continue with upload. If it does not, then I'd say implementation is inadequate.

mitjap:
@Murderlon I think this PR is now ready for some serious real world testing. 

On the side note: CI test on this PR is running tests against `master` branch.

Murderlon:
> @Murderlon I think this PR is now ready for some serious real world testing.

Perfect 👌 

> On the side note: CI test on this PR is running tests against `master` branch.

Fixed and merged into this branch. Tests are passing.

lib/handlers/PostHandler.js
```diff
@@ -1,46 +1,73 @@
...
         if ('upload-concat' in req.headers && !this.store.hasExtension('concatentation')) {
-            return Promise.resolve(super.send(res, 501, {}, 'Concatenation extension is not (yet) supported. Disable parallel uploads in the tus client. '));
+            throw ERRORS.UNSUPPORTED_CONCATENATION_EXTENSION;
+        }
+
+        const upload_length = req.headers['upload-length'];
+        const upload_defer_length = req.headers['upload-defer-length'];
+        const upload_metadata = req.headers['upload-metadata'];
+
+        if ((upload_length === undefined) === (upload_defer_length === undefined)) {
```
Murderlon, line 40:
```suggestion
        if (!upload_length && !upload_defer_length) {
```

lib/stores/FileStore.js
```diff
@@ -92,78 +75,73 @@ class FileStore extends DataStore {
...
+            flags: 'r+',
+            start: offset,
+        });
 
-            let new_offset = 0;
-            req.on('data', (buffer) => {
-                new_offset += buffer.length;
-            });
+        let new_offset = 0;
+        readable.pause();
```
Murderlon, line 126: why the `pause`/`resume` in the same function?

Murderlon (commented):
Some final comments. I think we are then ready to merge this.

lib/handlers/PostHandler.js
mitjap: This is not the same. Previous code does XOR and checks that either one is defined but not both.

lib/stores/FileStore.js
mitjap: Stream will switch in flowing mode when `data` listener is attached. This is to prevent any data loss before `stream.pipeline(...)` is called later on. But after some more thought I guess this is not needed as no data can be emited before execution reaches event loop.

lib/handlers/BaseHandler.js
```diff
@@ -31,14 +33,18 @@ class BaseHandler extends EventEmitter {
         return res.end();
     }
 
+    generateUrl(req, file_id) {
+        return this.options.relativeLocation ? `${req.baseUrl || ''}${this.options.path}/${file_id}` : `//${req.headers.host}${req.baseUrl || ''}${this.options.path}/${file_id}`;
```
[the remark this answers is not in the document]
mitjap: I think so. I have not changed this in any way.

lib/handlers/DeleteHandler.js
```diff
@@ -11,22 +11,16 @@ class DeleteHandler extends BaseHandler {
...
      * @return {function}
      */
-    send(req, res) {
+    async send(req, res) {
         const file_id = this.getFileIdFromRequest(req);
-        if (!file_id) {
-            console.warn('[DeleteHandler]: not a valid path');
-            return Promise.resolve(super.send(res, 404, {}, 'Invalid path name\n'));
+        if (file_id === false) {
+            throw ERRORS.INVALID_PATH;
```
Acconut, line 17: Other handles return ERRORS.INVALID_PATH if file_id === false. Should this be done here as well to keep it uniform?

lib/stores/DataStore.js
```diff
@@ -96,36 +60,25 @@ class DataStore extends EventEmitter {
...
-    getOffset(id) {
-        return new Promise((resolve, reject) => {
-            if (!id) {
-                return reject(ERRORS.FILE_NOT_FOUND);
-            }
-
-            return resolve({ size: 0, upload_length: 1 });
-        });
+    getOffset(file_id) {
+        return Promise.resolve({ size: 0, upload_length: 1 });
```
Acconut, line 81: Is there a reason for the non-zero upload length?

```suggestion
        return Promise.resolve({ size: 0, upload_length: 0 });
```

lib/stores/FileStore.js
```diff
@@ -45,45 +48,25 @@ class FileStore extends DataStore {
...
-                file_id = this.generateFileName(req);
-            }
-            catch (generateError) {
-                log('[FileStore] create: check your namingFunction. Error', generateError);
-                return reject(ERRORS.FILE_WRITE_ERROR);
-            }
-
-            const file = new File(file_id, upload_length, upload_defer_length, upload_metadata);
-            return fs.open(`${this.directory}/${file.id}`, 'w', (err, fd) => {
+            return fs.open(path.join(this.directory, file.id), 'w', async(err, fd) => {
```
Acconut, line 56: Rewriting this to use the fs Promise API from Node was not a goal of this PR, I assume?

lib/stores/FileStore.js
```diff
@@ -92,78 +75,73 @@ class FileStore extends DataStore {
...
+        const writeable = fs.createWriteStream(path.join(this.directory, file_id), {
+            flags: 'r+',
+            start: offset,
+        });
 
-            let new_offset = 0;
-            req.on('data', (buffer) => {
-                new_offset += buffer.length;
-            });
+        let new_offset = 0;
```
Acconut, line 125: The name of this variable is misleading. It does not track the new offset, but counts the bytes read. Maybe we should rename it?

lib/stores/FileStore.js
Acconut: Maybe we can also get rid of the data listener and simply stat the file again to see the new offset. Not sure if that's worth the effort though.

test/Test-FileStore.js
```diff
@@ -17,22 +24,63 @@ describe('FileStore', function () {
...
+      return this.datastore.create(file).should.be.rejected()
+    });
+
+    it('should resolve when the directory exists', function () {
+      return this.datastore.create(file).should.be.fulfilled();
+    });
+
+    it('should create an empty file', async function () {
+      // TODO: this test would pass even if `datastore.create` would not create any file
+      // as the file probably already exists from other tests
```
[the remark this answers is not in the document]
Acconut: Is this comment still relevant?

test/Test-Stores.shared.js
Acconut: I agree, it is probably best to not rely on `write` reporting the correct offset after an error.

index.d.ts
```diff
@@ -26,20 +36,22 @@ declare interface S3StoreOptions extends DataStoreOptions {
...
 
-declare class File {
+declare interface IFile {
     id: string;
-    upload_length: any;
-    upload_defer_length: any;
-    upload_metadata: any;
+    upload_length: string;
+    upload_defer_length: string;
+    upload_metadata: string;
```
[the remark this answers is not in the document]
mitjap: I'd love to. Maybe for next PR?

index.d.ts
```diff
@@ -2,17 +2,27 @@ import { EventEmitter } from "events";
 import * as fs from "fs";
 import * as http from "http";
 
+declare interface Configstore {
+    get(key: string) : Promise<IFile | undefined> | IFile | undefined;
```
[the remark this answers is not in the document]
mitjap: Neither am I. I'm not aware of optional return type in TypeScript. I went over the TypeScript documentation and could not find optional return type. Only optional parameters.

The idea is that Configstore can be synchronous or asynchronous and in both cases it can return `undefined` in case when there is no entry for specified key.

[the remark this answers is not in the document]
Murderlon: I think this type may be what you are looking for: https://github.com/sindresorhus/type-fest/blob/main/source/promisable.d.ts.

In the TypeScript rewrite I'll look more closely at creating the most correct types.

mitjap:
I'm not really sure why this last test is failing. It works for me with node 14 and 16. Also the code looks like it should be rejected as expected.
```
   1) S3DataStore
       write
         should reject when stream is destroyed:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/tus-node-server/tus-node-server/test/Test-S3DataStore.js)
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)
```

index.d.ts
```diff
@@ -50,22 +62,24 @@ export declare class DataStore extends EventEmitter {
...
 
 /**
  * file store in local storage
  */
 export declare class FileStore extends DataStore {
     constructor(options: FileStoreOptions);
-    read(file_id: string): fs.ReadStream;
-    getOffset(file_id: string): Promise<fs.Stats & File>;
+    read(file_id: string): stream.Readable;
+    getOffset(file_id: string): Promise<fs.Stats & IFile>;
```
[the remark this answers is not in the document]
Murderlon: Not for now, but it's on my todo list to make this consistent as currently it's inconsistent between stores

lib/stores/FileStore.js
Murderlon: One 1.0 hits we should put official supported node versions clearly in the readme.

test/Test-Stores.shared.js
```diff
@@ -1,224 +1,137 @@
...
+    it('should reject when stream is destroyed', async function () {
+      await this.datastore.create(file);
+
+      const readable = new stream.Readable();
+      const offset = this.datastore.write(readable, file.id, 0);
+
+      readable.push('some data');
+      readable.destroy();
+
+      return offset.should.be.rejected();
```
Murderlon, line 116: > I'm not really sure why this last test is failing. It works for me with node 14 and 16. Also the code looks like it should be rejected as expected.

I ran into this issue too with mocha sometimes, it's a bit annoying with `async`/`await`. I think what happens is that you don't await the `datastore.write`, so you may return an unresolved promise?

```suggestion
      await assert.rejects(offset, Error) // or something more specific than `Error`
```

Murderlon:
@mitjap any chance you'll have time for the final tiny steps to get this merged? I'm thinking of starting with the TypeScript rewrite and such after this PR to avoid merge conflicts.

test/Test-Stores.shared.js
mitjap: > @mitjap any chance you'll have time for the final tiny steps to get this merged? I'm thinking of starting with the TypeScript rewrite and such after this PR to avoid merge conflicts.
I will give this last test one more try to push this forward. If I can't fix this to work on CI I'll just comment out this test.

When you are on TypeScript rewrite please let me know if I can pitch in.

Murderlon:
I'm a bit confused why CI runs successfully on Node 16.x but not on 14.x? I don't see any errors about using something unsupported.

mitjap:
> 



> I'm a bit confused why CI runs successfully on Node 16.x but not on 14.x? I don't see any errors about using something unsupported.

I skipped GCS test and all other tests work.
https://github.com/tus/tus-node-server/actions/runs/3122672586/jobs/5064763568

Are credentials still valid? Is there some firewall preventing CI communication with GCS? Tests fail with `Retry limit exceeded` which is  thrown from `@google-cloud`.

Murderlon:
> Are credentials still valid? Is there some firewall preventing CI communication with GCS? Tests fail with `Retry limit exceeded` which is thrown from `@google-cloud`.

Testing locally with `npm run test:local`, which uses the unencrypted version of `keyfile.json.gpg` (same as in CI), works for me. This makes me think something did in fact change in this PR which makes the tests fail.

I'm very appreciative for your efforts in this PR but I think skipping all GCS tests is not an acceptable solution. I'll checkout this PR locally and see what I can do.

Murderlon:
The tests in this branch also run successfully for me locally and other PRs also still work the GCS tests 🤔. Bit of a mystery.

test/Test-GCSDataStore.js
```diff
@@ -1,12 +1,11 @@
 'use strict'
 const path = require('path')
 
-const Server = require('../lib/Server')
 const GCSDataStore = require('../lib/stores/GCSDataStore')
 
 const shared = require('./Test-Stores.shared')
 
-describe('GCSDataStore', () => {
+describe.skip('GCSDataStore', () => {
```
Murderlon, line 8:
```suggestion
describe('GCSDataStore', () => {
```

Murderlon:
Things work again 🤷‍♂️

Murderlon (approved):
Amazing work

mitjap:
I was not suggesting to leave test ignored, just wanted to see that other tests pass and that GCS is to blame. But I guess everything works as expected after some magic :)

Very happy to see this merged.

PR #34, opened 2016-05-31
Close event handler on request

sirakoff:
When upload is aborted, the writeStream doesn't end. So the file cannot be altered or deleted.

PR #188, opened 2021-06-09
Fix memory leaks on interrupted upload

mitjap:
When connection is closed by client (for example when pausing an upload), promise is never settled, which causes memory leak. This issue is present in FileStore and GCSDataStore (not sure about S3Store). In case of FileStore files might stay in write mode and locked.

When using `.pipe()` not all events are connected, so I switched it to new `stream.pipeline(source, destination, cb)`. 

This was inspired by #34.


Acconut:
Nice catch! i didn´t know `stream.pipe` could be dangerous and that there is a `stream.pipeline` method to work around this.

ErunamoJAZZ:
oh god, this would explain those memory leaks I got in a project sometime ago :S

Murderlon (approved):
Needs a merge from master but other than that this seems good to go

Murderlon:
Thanks!

PR #190, opened 2021-06-12
GET handler for serving uploaded files

mitjap:
This PR implements GET handler with previous behavior using registered handlers. Additionally also allows users to download files. This makes demo and uppy.io fully functional. Previously downloading a file that was just uploaded would return 404.

I understand this is not part of tus protocol but I think it makes better experience when using tus-node-server.


mitjap:
I found something simmilar in `tusd` implementation:
https://github.com/tus/tusd/blob/d560c4e7535cd21e6c9de5aaef335d2b159d8573/pkg/handler/unrouted_handler.go#L712

Minor difference is handling when data is not complete. My implementation returns 200 iff uploaded data (`Upload-Offset`) matches `Upload-Length`. In any other scenario it returns 404 meaning incomplete data.

Acconut:
Nice work on landing this feature!

PR #192, opened 2021-06-14
Add metadata parsing and stringification functions

mitjap:
Metadata is part of a standard and stores might or might not use it. I think that parsing and stringification should be implemented as part of this library (and properly test) even though it is not directly used as metadata is usually just stored in it's encoded form.

I use naming convention from `JSON.parse` and `JSON.stringify`. `Metadata.encode` and `Metadata.decode` can also be used.


Acconut:
:100: This will be very helpful!

PR #193, opened 2021-07-09
S3Store improvements

mitjap:
This PR addresses issues from https://github.com/tus/tus-node-server/issues/137. New implementation splits incoming request into multiple streams which are buffered to files and then uploaded when each part is at least 5MB. 

Parts that are smaller than minimum required size 5MB are discarded.


lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,173 @@
...
+
+    this.partial.bytes += data.length;
+    this.partial.stream.write(data);
+
+    if (overflow === 0) {
+      this.partial.stream.end();
+      this.partial = null;
+    }
+  }
+}
```
mitjap, line 60: I think this class should be removed. It was used as proof of concept.

package.json
```diff
@@ -56,8 +57,9 @@
     "tus-js-client": "^2.2.0"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.18.0",
     "@google-cloud/storage": "^5.3.0",
-    "aws-sdk": "^2.761.0",
+    "aws-sdk": "^2.929.0",
```
mitjap, line 62: Not sure which AWS SDK should be used. `aws-sdk` is used in the code but I think this is deprecated in favor of `@aws-sdk/client-s3˙. Need to re-check and remove one.

mitjap: I decided to remove `@aws-sdk/client-s3` as this is a complete rewrite and has diffrent API. This should be changed in separate PR.

cjmartin:
Any chance this is on track to be merged? I'm running into issues like those in #137 and it would be awesome to see this fix in.

Acconut:
@mitjap  Amazing work! :tada: Is this still a work in progress or already finished?

mitjap:
This implementation does not support patches smaller than 5MB which is minimum part size for Amazon S3. Other than that, I think it should work.

Acconut:
Great! In that case, I think we can merge and than work on support for PATCH requests smaller than 5MB in the future.

lib/models/StreamSplitter.js
Murderlon, line 60:
```suggestion
```

lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,173 @@
...
+        return resolve();
+      });
+    });
+  }
+}
+
+module.exports = {
+  FileStreamSplitter,
+  InMemoryStreamSplitter,
+};
```
Murderlon, line 173:
```suggestion
module.exports = { FileStreamSplitter };
```

lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,173 @@
...
+    this.filenameTemplate = randomString(10);
+
+    this.part = 0;
+
+    this.on('error', this._finishChunk.bind(this));
+
+  }
+
+  _write(chunk, encoding, callback) {
+    Promise.resolve()
```
Murderlon, line 82: I'm inclined to start using `async`/`await` everywhere and moving away from `bind`. I'll likely get to this in the TS rewrite, but it would already help if this big rewrite sets the pace for that. What do you think?

Murderlon (commented):
I'm not familiar enough with tus node to have a deep grasp on this, but it looks good as far as I can tell :)

Just to let you know I'm hoping to merge this and #188, after which I'm planning to tackle consistent formatting with prettier, proxy support, and the TS rewrite (all for the 1.0 release). I'm holding off on that for now to reduce merge conflicts. Speaking of which, there are a few still that need to be resolved.

And can we add some tests for this new functionality?

lib/models/StreamSplitter.js
mitjap: I reluctant to change anything at this points as I'm not able to test this thoroughly. I'm using GCS as my datastore. 
Otherwise I agree with async/await.

lib/models/StreamSplitter.js
mitjap: I recommend you release a 0.6 version prior to refactor as this might take some time to get to final version. I think tests for S3 should be made separately by refactoring some other tests and spliting them to general integration test that should work for all datastores and specific unit tests for each datastore.

lib/models/StreamSplitter.js
Murderlon: Good point. I'll make setting up a S3 test suite my next to do for tus-node-server. My time is limited on this project though, so will take a bit of time.

PR #86, opened 2017-12-31
Adds support for the termination extension

jclangst:
As defined: https://tus.io/protocols/resumable-upload.html#termination

Implementation is limited to a data store on the local file system and does not include support for the Google Cloud Services (yet)

Added the appropriate tests mirroring the existing test suite

PR #200, opened 2022-02-01
Add support for termination in FileStore

Murderlon:
Supersedes #86

> As defined: https://tus.io/protocols/resumable-upload.html#termination
>
> Implementation is limited to a data store on the local file system and does not include support for the Google Cloud Services (yet)
>
> Added the appropriate tests mirroring the existing test suite


Acconut:
Nice work!

PR #203, opened 2022-02-15
Add Metadata to types and run it through prettier

samal-rasmussen:
* Add Metadata
* Add typed return type to FileStore.getOffset
* Remove duplicated inherited methods from child stores that don't change the types they inherit from DataStore
* Make req parameter type on DataStore.create Partial
* Run through prettier for more standard formatting


Acconut (approved):
Thank you very much!

PR #206, opened 2022-02-21
Add configstore to FileStore type

samal-rasmussen:


samal-rasmussen:
I'm usning tus-node-server as express middleware. I am unable to use ~/.config in our production environment and it wouldn't survive a server redeploy anyways. So I've overridden the configstore with a simple in memory db:

```
const tusFileStore = new tus.FileStore({
  path: '/files',
})

const lru = new QuickLRU({
})

tusFileStore.configstore = {
  get(key: string) {
    return lru.get(key)
  },
  set(key: string, val: unknown) {
    lru.set(key, val)
  },
  delete(key: string) {
    lru.delete(key)
  },
}
```

As a solution with better encapsulation I would recommend that FileStore should take a db solution as a constructor argument, and fallback to configstore if non is provided.

Acconut:
@Murderlon Can you have a look at this?

Murderlon:
This is on my todo list, but not this type addition but instead a way to configure configstore more ergonomically

Acconut:
Sounds good! Thanks for the update!

samal-rasmussen:
Closing pr. Opening ticket #246 instead.

Issue #150, opened 2019-05-07
Race condition saving files to S3


DJWassink:
When saving using tus-node-server with the S3 backend the meta data gets saved in a `.info` file its metadata. 
This file also gets used to reply to a `HEAD` request which uppy uses to determine if an upload succeeded (which it does after a PATCH resulted in 502?). So there is a race condition where tus-node-server actually failed in saving the file (hence the 502) but the HEAD call returns that there is a file since tus-node-server only check the meta `.info` file.

Any idea how to solve this? Would an extra call to S3 to check if the main file also exists be a solution?

DJWassink:
Instead of an extra call to to check if the file actually exists, instead maybe it would be a better idea to write to the `.info` file with a flag notifying that the upload is completely done? This could happend in the `_finishMultipartUpload` method:

https://github.com/tus/tus-node-server/blob/4b44018d1cb1cc7c0c5c770f96581ef28c3cf1d9/lib/stores/S3Store.js#L330-L350

DJWassink:
After some more debugging it seems like the there exists a case that the `_finishMultipartUpload` never gets called but the part is uploaded. Thus the meta data call will think that all parts are uploaded but the multipart upload never actually finished.

No idea how to fix this. For now I have made a clone of `tus-node-server` and also check if there is an ongoing multipart upload ongoing with the current file_id:

```js
    isMultipartUploadFinished(file_id) {
        return this.client.listMultipartUploads({
            Bucket: this.bucket_name,
            Prefix: file_id
        }).promise().then((data) => {
            return data.Uploads.some(upload => upload.Key === file_id) === false;
        }).catch((err) => {
            throw err;
        })
    }
```

This piece of code is added to the S3Store which I call from the HeadHandler which will return a 404 if the mutlipart upload is still ongoing resulting in Uppy creating a complete new upload. Sadly this is a very dirty solution since S3 will keep the previous upload reserved thus we get chared for "unused" storage.

Acconut:
As far as I know, tus-node-server has little protection against concurrent access which can result in these race conditions as you mentioned. I also don't know how to solve this but we're happy to accept a PR if you could supply one.

Alternatively, I would recommend you to consider using another tus server (such as the more mature https://github.com/tus/tusd) as we tus-node-server is in a bad shape right now (see https://github.com/tus/tus-node-server/blob/master/README.md).

DJWassink:
[...]

PR #207, opened 2022-02-24
Fix for HEAD request not returning Upload-Offset when S3 multipart upload is finished

mzoellner:
Hi,
we recently ran into something quite similar as reported in #150.

We are using tus-node-server with the S3 store in combination with @uppy/tus and occasionally we see uploads that never finish on the client side, but the file has been successfully written to S3. The client retries the upload a couple times, but it can't and eventually fails.

We suspect network issues to be the culprit of this, but it's hard to say because we didn't really have our client and server instrumented with enough debug logs to tell.

What we believe is happening is as follows:
When the last part of an upload is finished (and the file is successfully written to and concatenated in S3), the response of this last PATCH request does not necessarily end up on the client side (network errors or cancelled requests may be the reason but it's hard to tell). @uppy/tus then tries to find out about the state of the upload with a HEAD request, but this returns a 404 because the multipart upload does not exist anymore and when tus-node-server tries to retrieve the upload parts we get an error from S3.

I managed to reproduce this scenario, by throwing an error in tus-node-server, just after the upload was completed in s3 and before the response was sent back to the client.

To fix the issue, I made sure that the HEAD request returns the correct response, even if the upload itself is already completed. It does so by checking if the file exists in S3 and if so, returns the correct Upload-Offset header.
This in turn satisfies @uppy/tus and the file is correctly marked as uploaded.

If I understand the [protocol](https://tus.io/protocols/resumable-upload.html#head) correctly it states that a HEAD request should always return the Upload-Offset header in its response, even if the upload is already completed:

> The Server MUST always include the Upload-Offset header in the response for a HEAD request, even if the offset is 0, or the upload is already considered completed.

So this fix is actually making sure that tus-node-server with S3 store is correctly adhering to the TUS protocol.

It would be great if you could consider merging this.
Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,21 +192,14 @@ server.get('/uploads', (req, res) => {
 ```
 
 #### Custom file names:
-
-The default naming of files is a random crypto hex string. When using your own `namingFunction`, make sure to create URL friendly names such as removing spaces.
-
 ```js
-const crypto = require('crypto');
-
-// req is http.IncomingMessage
-const randomString = (req) => {
-    // same as the default implementation
-    return crypto.randomBytes(16).toString('hex');
+const fileNameFromUrl = (req) => {
+    return req.url.replace(/\//g, '-');
 }
 
 server.datastore = new tus.FileStore({
     path: '/files',
-    namingFunction: randomString
+    namingFunction: fileNameFromUrl
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -192,14 +192,21 @@ server.get('/uploads', (req, res) => {
 ```
 
 #### Custom file names:
+
+The default naming of files is a random crypto hex string. When using your own `namingFunction`, make sure to create URL friendly names such as removing spaces.
+
 ```js
-const fileNameFromUrl = (req) => {
-    return req.url.replace(/\//g, '-');
+const crypto = require('crypto');
+
+// req is http.IncomingMessage
+const randomString = (req) => {
+    // same as the default implementation
+    return crypto.randomBytes(16).toString('hex');
 }
 
 server.datastore = new tus.FileStore({
     path: '/files',
-    namingFunction: fileNameFromUrl
+    namingFunction: randomString
 });
 ```
 
--- a/lib/stores/S3Store.js
+++ b/lib/stores/S3Store.js
@@ -500,6 +500,21 @@ class S3Store extends DataStore {
         });
     }
 
+    isUploadFinished(file_id) {
+        return this.client
+            .headObject({
+                Bucket: this.bucket_name,
+                Key: file_id,
+            })
+            .promise()
+            .then((data) => {
+                return true
+            })
+            .catch((err) => {
+                return false
+            });
+    }
+
     getOffset(file_id, with_parts = false) {
         return new Promise((resolve, reject) => {
             this._getMetadata(file_id)
@@ -512,10 +527,25 @@ class S3Store extends DataStore {
                             };
                         })
                         .catch((err) => {
-                            throw err;
+                            if (with_parts === false) {
+                                return this.isUploadFinished(file_id).then(isFinished => {
+                                    if (isFinished) {
+                                        return { metadata, isFinished };
+                                    }
+                                    throw err
+                                })
+                            }
                         });
                 })
                 .then((data) => {
+                    if (data.isFinished) {
+                        let output = Object.assign({}, this.cache[file_id].file, {
+                            size: data.metadata.file.upload_length, 
+                            upload_length: data.metadata.file.upload_length
+                        });
+                        return resolve(output)
+                    }
+
                     // if no parts are found, offset is 0
                     if (data.parts.length === 0) {
                         return resolve({
--- a/lib/stores/S3Store.js
+++ b/lib/stores/S3Store.js
@@ -520,12 +520,14 @@ class S3Store extends DataStore {
                             return this.client.headObject({ Bucket: this.bucket_name, Key: file_id })
                                 .promise()
                                 .then(() => {
-                                    resolve({
-                                        ...this.cache[file_id].file,
-                                        size: metadata.file.upload_length,
-                                        upload_length: metadata.file.upload_length,
-                                    });
-                                    return { isFinished: true };
+                                    return {
+                                        data: {
+                                            ...this.cache[file_id].file,
+                                            size: metadata.file.upload_length,
+                                            upload_length: metadata.file.upload_length,
+                                        },
+                                        isFinished: true,
+                                    };
                                 })
                                 .catch((headError) => {
                                     throw headError;
@@ -534,15 +536,14 @@ class S3Store extends DataStore {
                 })
                 .then((data) => {
                     if (data.isFinished) {
-                        return;
+                        return resolve(data.data);
                     }
                     // if no parts are found, offset is 0
                     if (data.parts.length === 0) {
-                        resolve({
+                        return resolve({
                             size: 0,
                             upload_length: data.metadata.file.upload_length,
                         });
-                        return;
                     }
 
                     const offset = data.parts.reduce((a, b) => {
@@ -559,8 +560,7 @@ class S3Store extends DataStore {
                         output = Object.assign(output, { parts: data.parts });
                     }
 
-                    resolve(output);
-                    return;
+                    return resolve(output);
                 })
                 .catch((err) => {
                     if (['NotFound', 'NoSuchUpload'].includes(err.code)) {
--- a/lib/stores/S3Store.js
+++ b/lib/stores/S3Store.js
@@ -508,10 +508,10 @@ class S3Store extends DataStore {
             })
             .promise()
             .then((data) => {
-                return true
+                return true;
             })
             .catch((err) => {
-                return false
+                return false;
             });
     }
 
@@ -528,22 +528,23 @@ class S3Store extends DataStore {
                         })
                         .catch((err) => {
                             if (with_parts === false) {
-                                return this.isUploadFinished(file_id).then(isFinished => {
+                                return this.isUploadFinished(file_id).then((isFinished) => {
                                     if (isFinished) {
                                         return { metadata, isFinished };
                                     }
-                                    throw err
-                                })
+                                    throw err;
+                                });
                             }
+                            throw err;
                         });
                 })
                 .then((data) => {
                     if (data.isFinished) {
-                        let output = Object.assign({}, this.cache[file_id].file, {
-                            size: data.metadata.file.upload_length, 
-                            upload_length: data.metadata.file.upload_length
+                        const output = Object.assign({}, this.cache[file_id].file, {
+                            size: data.metadata.file.upload_length,
+                            upload_length: data.metadata.file.upload_length,
                         });
-                        return resolve(output)
+                        return resolve(output);
                     }
 
                     // if no parts are found, offset is 0
--- a/lib/stores/S3Store.js
+++ b/lib/stores/S3Store.js
@@ -520,11 +520,12 @@ class S3Store extends DataStore {
                             return this.client.headObject({ Bucket: this.bucket_name, Key: file_id })
                                 .promise()
                                 .then(() => {
-                                    return resolve({
+                                    resolve({
                                         ...this.cache[file_id].file,
                                         size: metadata.file.upload_length,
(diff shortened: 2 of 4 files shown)


lib/stores/S3Store.js
```diff
@@ -512,10 +527,25 @@ class S3Store extends DataStore {
                             };
                         })
                         .catch((err) => {
-                            throw err;
+                            if (with_parts === false) {
```
mzoellner, line 530: This if statement is a bit awkward, but `getOffset` is also called from within the s3 store with `with_parts: true` and I wanted to avoid any side effects.

mitjap:
I recommend you take a look at PR https://github.com/tus/tus-node-server/pull/193. It fixes many of the issues with current S3 store. Might also fix yours. It is in "draft" because but I think it should work. This PR is planned to be included in 1.0.0. version.

mzoellner:
I had a look at your PR @mitjap, but I can't immediately see how it would change the incorrect behavior of the HEAD request. The protocol states that a HEAD request should always return the Upload-Offset header and I don't think your PR is fixing that (correct me if I'm wrong).

I'll nonetheless have a look at it and try to see if it fixes the issue.

Is there any indication as to when 1.0.0 is landing ?

Murderlon (changes requested):
Thanks for the contribution but I'm not sure about this implementation (yet)

- Making a head request when `._retrieveParts` fails for whatever reason seems excessive. Can this be more specific to a certain error?
- I'm unsure about the `if (with_parts === false)`. Why is it required?
- How can I test this works? Ideally we would write a test, but as we currently don't have any tests for S3 I don't think I can ask that of you.
- There are quite a few eslint errors
- I have issues with some of the code, like making a new class method that is only used once, and the fact that it is spread over three places. I think it could be all in one place.

Murderlon:
> Is there any indication as to when 1.0.0 is landing ?

I'm starting work on that as of now but there is no clear timeline. Quite a few big things need to land first.

mzoellner:
Hi @Murderlon ,
thank you for taking the time to respond.

I'll try to answer your questions:

> There are quite a few eslint errors
Sorry about that. I missed this completely since my editor did somehow not show them. Should be fixed now.

> Making a head request when ._retrieveParts fails for whatever reason seems excessive. Can this be more specific to a certain error?
Totally right. I changed it so it only makes a HEAD request when the error is "NoSuchUpload" which indicates that either there was no upload for this `fileId` or it has already completed (see S3 docs [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)). I don't think we can make it more specific than this and frankly I think one small HEAD request in an error case is justified, considering that a user could have uploaded gigabytes only to then have the upload fail at the last request.

> I'm unsure about the if (with_parts === false). Why is it required?
Yeah this is ugly and I don't like it either. `getOffset` is called from both the HEAD request and the PATCH request (via the `_processUpload` method). The HEAD request calls it (implicitly) with `with_parts: false`, the `_processUpload` method calls it (explicitly) with `with_parts: true`. The `_processUpload` method expects upload parts to be returned, which would not work when the upload is completed, since there are no parts anymore. The smallest change therefore was to use this if statement to make sure we are in the HEAD request code path. We could instead pass in another flag to the `getOffset` method to separate the two concerns, e.g. `checkIfUploadIsFinished`, or do a bigger refactoring. What do you reckon ?

> I have issues with some of the code, like making a new class method that is only used once, and the fact that it is spread over three places. I think it could be all in one place.
I am not sure I understand your concern. What is the problem with a class method that is used only once ? Performance ?
I had a few iterations on this and tried to limit it to as few locations as possible, but the lack of tests (and types) makes it really hard to not break something else. You might be more familiar with the codebase than I am. Feel free to suggest another way of fixing this.

> How can I test this works? Ideally we would write a test, but as we currently don't have any tests for S3 I don't think I can ask that of you.
If you provide a bare bones test harness for S3 (i.e. a bucket I can write into), I am more than happy to write an integration test. It should be quite straightforward to test actually:
- Create an upload and upload the entire file
- Make a HEAD request to the same fileId
- Assert that the HEAD request is returning 200 instead of 404 and the correct Upload-Offset header

Sorry for the long text. I really appreciate you taking time for this. I am not in a hurry to get this in. Let's get it right first.

On a side note: We have this fix running in production for a week now and haven't seen any issues anymore, whereas we had multiple issues before this every day.

Thank you

Murderlon:
I couldn't figure out how to merge the latest master from origin to your forked branch. Currently the latest changes in master are undone in this PR. Could you merge latest master for me?

lib/stores/S3Store.js
```diff
@@ -505,14 +505,30 @@ class S3Store extends DataStore {
...
                         .catch((err) => {
-                            throw err;
+                            // getOffset is called from both the HEAD request and the PATCH request
+                            // (via the _processUpload method). The HEAD request calls it (implicitly)
+                            // with with_parts: false, the _processUpload method calls it (explicitly)
+                            // with with_parts: true. The _processUpload method expects upload parts to be returned,
+                            // which would not work when the upload is completed, since there are no parts anymore.
+                            // TODO: refactor to separate these concerns.
+                            // TODO: write a test for this case.
+                            if (with_parts === true || err !== 'NoSuchUpload') {
```
Murderlon, line 517: One thing I'm not sure about is whether we can compare `err` with a string. If it an `Error` object, this won't work. (I'm rewriting this to TS in the nearby future with should help with these silly things).

Murderlon:
Seems my commit with changes got removed: https://github.com/tus/tus-node-server/commit/082e8b91d15ed3f86a8db8267f9e002a8fece60c. Can we try to bring those changes back?

lib/stores/S3Store.js
mzoellner: I don't think this will work, since the error returned is an object that contains a property `code` with the value 'NoSuchUpload`

Murderlon, line 517:
```suggestion
                            if (with_parts === true || err.code !== 'NoSuchUpload') {
```

lib/stores/S3Store.js
```diff
@@ -505,14 +505,30 @@ class S3Store extends DataStore {
...
+                                throw err;
+                            }
+                            return this.client.headObject({ Bucket: this.bucket_name, Key: file_id })
+                                .promise()
+                                .then((data) => {
+                                    return resolve({
+                                        ...this.cache[file_id].file,
+                                        size: data.metadata.file.upload_length,
+                                        upload_length: data.metadata.file.upload_length,
```
Murderlon, line 526:
```suggestion
                                .then(() => {
                                    return resolve({
                                        ...this.cache[file_id].file,
                                        size: metadata.file.upload_length,
                                        upload_length: metadata.file.upload_length,
```

README.md
```diff
@@ -192,21 +192,14 @@ server.get('/uploads', (req, res) => {
...
-The default naming of files is a random crypto hex string. When using your own `namingFunction`, make sure to create URL friendly names such as removing spaces.
-
 ```js
-const crypto = require('crypto');
-
-// req is http.IncomingMessage
-const randomString = (req) => {
-    // same as the default implementation
-    return crypto.randomBytes(16).toString('hex');
+const fileNameFromUrl = (req) => {
```
Murderlon, line 196: these changes should not happen here. These seem to undo latest master

mzoellner: ouch. I just cherry-picked your commit. let me fix this

Murderlon (commented):
It seems the update to the eslintrc was reverted. That was needed to not have warnings for TODO comments.

mzoellner:
Sorry about the mess up. I cherry picked your commit, but that was outdated, so it reverted back the changes on master. It's all fixed now

lib/stores/S3Store.js
```diff
@@ -505,14 +505,30 @@ class S3Store extends DataStore {
...
+                                    });
+                                })
+                                .catch(() => {
+                                    throw err;
+                                });
```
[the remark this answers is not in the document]
mzoellner: yeah. that makes sense

Murderlon:
> I have potentially found one issue with your changes.

Yes feel free to double check if this works

Murderlon:
> I have potentially found one issue with your changes.

Did you find the issue in the end?

lib/stores/S3Store.js
```diff
@@ -505,14 +505,30 @@ class S3Store extends DataStore {
...
+                            // which would not work when the upload is completed, since there are no parts anymore.
+                            // TODO: refactor to separate these concerns.
+                            // TODO: write a test for this case.
+                            if (with_parts === true || err.code !== 'NoSuchUpload') {
+                                throw err;
+                            }
+                            return this.client.headObject({ Bucket: this.bucket_name, Key: file_id })
+                                .promise()
+                                .then(() => {
+                                    return resolve({
```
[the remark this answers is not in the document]
mzoellner: ouch sorry @Murderlon. I did not see that this was not submitted. I am not familiar with the github workflows yet.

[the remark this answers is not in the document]
Murderlon: Thanks for the effort, you're right. This code could use a refactor but for now we can indeed pass an extra `isFinished` in the `resolve` and return it again right away in the other `then`. 

We should also add this again for the TODO warnings: 
https://github.com/tus/tus-node-server/blob/082e8b91d15ed3f86a8db8267f9e002a8fece60c/.eslintrc#L24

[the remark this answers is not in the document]
mzoellner: I can make those changes.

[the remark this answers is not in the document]
mzoellner: The changes are in @Murderlon 
I had to change the returns in the `then` block, otherwise eslint wouldn't like it (consistent returns). Shouldn't be an issue though since `resolve()` returns `void` anyway.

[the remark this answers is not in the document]
mzoellner: I'll do a round of testing now, just to ensure nothing is broken. I don't like writing this without any type checks or tests...

lib/stores/S3Store.js
```diff
@@ -505,23 +505,44 @@ class S3Store extends DataStore {
...
+                            }
+                            return this.client.headObject({ Bucket: this.bucket_name, Key: file_id })
+                                .promise()
+                                .then(() => {
+                                    resolve({
+                                        ...this.cache[file_id].file,
+                                        size: metadata.file.upload_length,
+                                        upload_length: metadata.file.upload_length,
+                                    });
+                                    return { isFinished: true };
```
Murderlon, line 528: resolve and return mixed seems a bit strange
```suggestion
                                   return {
                                        data: {
                                            ...this.cache[file_id].file,
                                            size: metadata.file.upload_length,
                                            upload_length: metadata.file.upload_length,
                                        },
                                        isFinished: true,
                                    };
```

lib/stores/S3Store.js
```diff
@@ -505,23 +505,44 @@ class S3Store extends DataStore {
...
+                                })
+                                .catch((headError) => {
+                                    throw headError;
+                                });
                         });
                 })
                 .then((data) => {
+                    if (data.isFinished) {
+                        return;
+                    }
```
Murderlon, line 538: Together with the above change. (Also I think it's fine to `return resolve`, doesn't necessarily need to be separate).
```suggestion
                .then(({ data, isFinished }) => {
                    if (isFinished) {
                        return resolve(data);
                    }
```

lib/stores/S3Store.js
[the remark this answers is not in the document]
Murderlon: I know, next up for me is setting up a test suite for S3 and rewriting to TS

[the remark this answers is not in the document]
mzoellner: awesome

lib/stores/S3Store.js
mzoellner: yeah agree.
now it's pretty similar to where we started from :-)
I'll change it

mzoellner: ok. that's in now. I am doing more testing now

mzoellner: I have finished my testing with the current state and haven't seen any issues. I simulated the error case and I tested the default case.

So from my side this is all good to go.

Murderlon (approved):

Murderlon:
Thanks a lot @mzoellner!

mzoellner:
No thank you @Murderlon 
I appreciate you taking time to do this.

PR #209, opened 2022-03-24
Rewrite all tests for stores & share tests between stores

Murderlon:
- Create `Test-S3Store.js`
- Create `Test-Stores.shared.js`
- Reuse tests between file store, S3, and GCS
- Rewrite tests to not depend on each other. Previously tests depended on state being set to the server in a previous test. 
- Add secrets to GitHub Actions
- Add local testing command with `env-cmd` which consumes a local `.env` (I tried `dotenv` but `mocha` didn't like it)
- Fix race condition in `write` function of S3 store surfaced by tests
- Move output of tests from `/files` to `/test/output`
- Move files used for tests to `/test/fixtures`
- Format the test files with Prettier. (Is it consistent with the rest of the codebase? No. But it's inconsistent already and it doesn't matter as I will rewrite it to TS with Prettier anyway)


Murderlon:
friendly ping @Acconut

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+
+    it('should delete the file when it does exist', async function () {
+      const fileDeletedEvent = sinon.fake()
+      const req = { headers: { 'upload-length': 1000 }, url: this.storePath }
+
+      this.server.datastore.on(EVENTS.EVENT_FILE_DELETED, fileDeletedEvent)
+
+      const file = await this.server.datastore.create(req)
+      await this.server.datastore.remove({ file_id: file.id })
+      assert.equal(fileDeletedEvent.calledOnce, true)
```
Acconut, line 104: This test ensures that the event is correctly fired. Is it also possible to check whether subsequent calls for the deleted file returns errors?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+}
+
+exports.shouldWriteUploads = function () {
+  describe('write', function () {
+    it('should reject write streams that cant be opened', function () {
+      const stream = fs.createReadStream(this.testFilePath)
+      return this.server.datastore.write(stream, null, 0).should.be.rejectedWith(500)
+    })
+
+    xit('should open a stream, resolve the new offset, and emit upload complete', function (done) {
```
Acconut, line 116: Is the xit intended here?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+          .then((offset) => {
+            assert.equal(offset, size)
+            assert.equal(uploadCompleteEvent.calledOnce, true)
+            return done()
+          })
+          .catch(done)
+      })
+    })
+
+    xit('should settle on closed input stream', function (done) {
```
Acconut, line 137: Is the xit intended here?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+      await this.server.datastore.remove({ file_id: file.id })
+      assert.equal(fileDeletedEvent.calledOnce, true)
+    })
+  })
+}
+
+exports.shouldWriteUploads = function () {
+  describe('write', function () {
+    it('should reject write streams that cant be opened', function () {
+      const stream = fs.createReadStream(this.testFilePath)
```
Acconut, line 112: Can you explain why the stream could not be opened here? It looks fine for me.

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+      const uploadCompleteEvent = sinon.fake()
+
+      this.server.datastore.on(EVENTS.EVENT_UPLOAD_COMPLETE, uploadCompleteEvent)
+
+      const stream = fs.createReadStream(this.testFilePath)
+      const name = this.testFileName
+      const size = this.testFileSize
+
+      stream.once('open', () => {
+        this.server.datastore
```
Acconut, line 126: Can you explain where the upload for this write call is created? In a previous test method?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+
+      stream.pause()
+      stream.on('data', () => stream.destroy())
+
+      this.server.datastore
+        .create(req)
+        .then((file) => {
+          return this.server.datastore.write(stream, file.id, 0)
+        })
+        .catch(() => {})
```
Acconut, line 156: Is an error thrown in this case that we ignored on purpose?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+
+    it('should reject directories', function () {
+      return this.server.datastore.getOffset('').should.be.rejectedWith(404)
+    })
+
+    it('should resolve the stats for existant files', function () {
+      const req = { headers: { 'upload-length': this.testFileSize } }
+
+      this.server.datastore.create(req).then((file) => {
+        this.server.datastore.getOffset(file.id).should.be.fulfilledWith(this.testFileSize)
```
Acconut, line 176: I don't fully understand this test. We create an upload of the size `this.testFileSize` and then we obtain the upload offset without uploading anything. Shouldn't the offset by 0 instead of `this.testFileSize` in this case?

test/Test-Stores.shared.js
```diff
@@ -0,0 +1,180 @@
...
+      const name = this.testFileName
+      const size = this.testFileSize
+
+      stream.once('open', () => {
+        this.server.datastore
+          .write(stream, name, 0)
+          .then((offset) => {
+            assert.equal(offset, size)
+            assert.equal(uploadCompleteEvent.calledOnce, true)
+            return done()
```
Acconut, line 131: Can we also add an assertion that getOffset returns the new and correct offset?

Acconut (commented):
Awesome work, thank you very much! It looks mostly good but I have a few questions.

test/Test-Stores.shared.js
Murderlon: The stream is not ready yet, you need to wait for `stream.once('open')`

test/Test-Stores.shared.js
Murderlon: As stated in the todo's, I can't get this to work for S3 and I don't know why

test/Test-Stores.shared.js
Murderlon: I didn't write this test and I'm not really sure what this is testing because there is an error that would make the test fail if you put done in catch. I suppose it wants to see if the chain continues even if there is an error?

test/Test-Stores.shared.js
Murderlon: AFAIK `create` already fully uploads the file so the offset is the file size.

PR #244, opened 2022-05-05
Use tus-node-server with Fastify

### LINKED ISSUE #241: TIP: use as fastify plugin (no middleware)
Not an issue but a hint (maybe could be added to readme, but hey commit a PR for some words?)

simple mvp/poc looks like:

```js
// ie. ./services/upload.js
'use strict'

const path = require('path')
const tus = require('tus-node-server')

module.exports = (fastify, opts, next) => {
  const tusServer = new tus.Server()

  tusServer.datastore = new tus.FileStore({
    path: '/services/media/upload', // <-- full url path (ie.: fastify-plugin serving '/uploads' prefixed by '/services/media')
    directory: path.join(__dirname, '..', 'files') // <-- full filesystem path (ie.: './files')
  })

  fastify.addContentTypeParser(
    'application/offset+octet-stream',
    (request, body, done) => done()
  )

  fastify.all('/upload', (req, res) => {
    tusServer.handle(req.raw, res.raw)
  })

  fastify.all('/upload/:idUpload', (req, res) => {
    tusServer.handle(req.raw, res.raw)
  })

  next()
}

```

So if you wonder how to integrate tus support within fastify as plugin, this is a starting point. Keys are: 

- tus uses raw nodejs `req` and `res` - so: use them for handler `(req.raw, res.raw)`
- fastify defaults to _limit_ handling application/json - so: extend to tus' offset octet stream

That's it. Short, readable, customizable, IMOHO :)

mashpie:
This adds a standalone example to README. Closes #241 
I tried to align with the other examples and code style.

> ...and explain as much as possible?

Well, I tried to focus on relevant aspects only without commenting too much, to get a "Quick Start" setup. There is more, like Authentication, Routing, different Storages (esp. their settings) - which is better document in each projects docs.


Murderlon (approved):

Murderlon:
Thanks a lot

mashpie:
Happy tus user, happy to help :)

PR #245, opened 2022-05-05
Implement file stream splitter for S3

### LINKED ISSUE #137: Multipart Uploads to S3 fail

Murderlon:
Supersedes #193 
Closes #137

@mitjap creating a new PR so we both have access to the PR (I couldn't push to yours).

Since #209 is in we can now be confident to work on this and merge it.


lib/stores/S3Store.js
```diff
@@ -276,57 +286,96 @@ class S3Store extends DataStore {
...
+                    //     const stats = await fs.promises.stat(filepath);
+                    //     if (!isFinalRequest) {
+                    //         // We could modify chunk size in a way that last part is as small as possible.
+                    //         // Ideally we would not discard any data.
+                    //         if (stats.size < 5 * 1024 * 1024) {
+                    //             throw new Error('Part too small');
+                    //         }
+                    //     }
+                    //     return undefined;
+                    // })
```
Murderlon, line 353: @mitjap I had to remove this check. You were accessing headers from `req` but in all stores and tests `req` is a stream, not a http request, so the headers don't exist. Need to find a different way to handle this

lib/stores/S3Store.js
```diff
@@ -438,138 +500,141 @@ class S3Store extends DataStore {
...
+                                        size: metadata.file.upload_length,
+                                        upload_length: metadata.file.upload_length,
+                                    },
+                                    isFinished: true,
+                                };
+                            })
+                            .catch((headError) => {
+                                throw headError;
+                            });
+                    });
```
Murderlon, line 600: @mitjap this whole section was implemented after your initial version of this PR in #207. However, since you're rewrite this is no longer conditionally called with `with_parts` so we can remove the argument and clean this up

Murderlon (commented):
@mitjap I reckon we should now create some specific tests for this as well

lib/stores/S3Store.js
Murderlon: I was initially implementing an alternative way to check this and error but then then I realized this is redundant. It should never happen that a chunk is smaller than 5mb if it's not the final chunk (if a file is bigger than 5mb) and we could test that. I also don't think tusd errors on this?

Perhaps related:
https://github.com/tus/tusd/blob/d72854fa048e314a86b60cad98c8144341f297b5/pkg/s3store/s3store.go#L708-L717

lib/stores/S3Store.js
Murderlon: Solved in https://github.com/tus/tus-node-server/pull/245/commits/28ced0409b94cf4495a282da90f3d4f13298f2d5

lib/stores/S3Store.js
```diff
@@ -75,21 +68,28 @@ class S3Store extends DataStore {
...
 
     /**
-     * Check if the bucket exists in S3.
-     *
-     * @return {Promise}
-     */
+   * Check if the bucket exists in S3.
+   *
+   * @return {Promise}
+   */
```
Acconut, line 89: Regarding the formatting: The comment's indentation is one space less than for the method below. This looks a bit odd. Is this intended and enforced by the linter? Not a big deal, just wondering.

lib/stores/S3Store.js
```diff
@@ -276,57 +276,81 @@ class S3Store extends DataStore {
...
+   * @param {http<IncomingMessage>} req                 incoming request
+   * @param {Number}                currentPartNumber number of the current part/chunk
+   * @return {Promise<Number>}                         which resolves with the current offset
+   * @memberof S3Store
+   */
+    _processUpload(metadata, req, currentPartNumber) {
+        return new Promise((resolve, reject) => {
+            const splitter = new FileStreamSplitter(
+                {},
+                { maxChunkSize: this.part_size, directory: '.' }
```
Acconut, line 295: > directory: '.'

Does this mean that the temporary chunk files will be written in the current working directory? Would it be better to save them directly in the system's default temporary directory?

lib/stores/S3Store.js
```diff
@@ -276,57 +276,81 @@ class S3Store extends DataStore {
...
+            );
+
+            const promises = [];
+
+            let pendingChunkFilepath = null;
+
+            stream.pipeline(req, splitter, (pipelineErr) => {
```
Acconut, line 302: Is `stream.pipeline` a better version of the `ReadbleStream#pipe` method?

lib/stores/S3Store.js
```diff
@@ -276,57 +276,81 @@ class S3Store extends DataStore {
...
+
+                const p = Promise.resolve()
+                    .then(() => {
+                        return this._uploadPart(metadata, fs.createReadStream(filepath), partNumber);
+                    })
+
+                    .finally(() => {
+                        fs.rm(filepath, (err) => {
+                            if (err) {
+                                console.error(err);
```
Acconut, line 336: Can we also have a proper log message here, like:

```js
 log(`[${metadata.file.id}] failed to remove chunk ${pendingChunkFilepath}`);
```

lib/stores/S3Store.js
```diff
@@ -276,57 +276,81 @@ class S3Store extends DataStore {
...
+                    }
                 }
 
-                return current_offset.size;
+                promises.push(pipelineErr ? Promise.reject(pipelineErr) : Promise.resolve());
+                resolve(promises);
+            });
+
+            splitter.on('chunkStarted', (filepath) => {
+                pendingChunkFilepath = filepath;
```
Acconut, line 320: pendingChunkFilepath always points to the file that the FileStreamSplitter current writes to, doesn't it? So it is never possible to have two chunkStarted events after another without one chunkFinished in between?

lib/stores/S3Store.js
```diff
@@ -276,57 +276,81 @@ class S3Store extends DataStore {
...
+            });
+
+            splitter.on('chunkFinished', (filepath) => {
+                pendingChunkFilepath = null;
+
+                const partNumber = currentPartNumber++;
+
+                const p = Promise.resolve()
+                    .then(() => {
+                        return this._uploadPart(metadata, fs.createReadStream(filepath), partNumber);
```
Acconut, line 330: Does _uploadPart take care of closing the read stream properly?

lib/stores/S3Store.js
```diff
@@ -373,31 +397,44 @@ class S3Store extends DataStore {
...
                 return data.Parts;
+            })
+            .then((parts) => {
+                // sort and filter only for call where `part_number_marker` is not set
+                if (part_number_marker === undefined) {
+                    parts.sort((a, b) => {
+                        return a.PartNumber - b.PartNumber;
+                    });
+
+                    parts = parts.filter((value, index) => {
```
Acconut, line 413: Why is this filtering necessary? It ensures that we have a consecutive number of parts, but is this important? Can't I have part numbers 1, 2, 3, 5, 6?

lib/stores/S3Store.js
```diff
@@ -437,139 +474,107 @@ class S3Store extends DataStore {
...
-
-                                return this.getOffset(file_id, true)
-                                    .then((current_offset) => resolve(current_offset.size))
-                                    .catch(reject);
+    write(req, file_id) {
+        return this._getMetadata(file_id)
+            .then((metadata) =>
+                this._countParts(file_id).then((part_number) => [part_number, metadata])
+            )
+            .then(async(results) => {
```
Acconut, line 482: Can we have a space here or does the linter not allow this:

> async (results) => {

lib/stores/S3Store.js
```diff
@@ -437,139 +474,107 @@ class S3Store extends DataStore {
...
+                                            location,
+                                        }),
+                                    });
+
+                                    this._clearCache(metadata.file.id);
+
+                                    return current_offset.size;
+                                })
+                                .catch((err) => {
+                                    console.error(err);
```
Acconut, line 507: Should we have better error logging here?

lib/stores/S3Store.js
```diff
@@ -437,139 +474,107 @@ class S3Store extends DataStore {
...
+                            return this.getOffset(file_id).then((current_offset) => current_offset.size);
+                        }
 
-                        return a;
-                    }, 0);
+                        this._clearCache(file_id);
 
-                    let output = Object.assign({}, this.cache[file_id].file, {
-                        size: offset,
+                        console.error(err);
```
Acconut, line 535: Should we have better error logging here?

lib/stores/S3Store.js
```diff
@@ -437,139 +474,107 @@ class S3Store extends DataStore {
...
+            };
+        }
+        catch (err) {
+            if (err.code !== 'NoSuchUpload') {
+                console.error(err);
+                throw err;
+            }
+            // When the last part of an upload is finished and the file is successfully written to S3,
+            // the upload will no longer be present and requesting it will result in a 404.
+            // In that case we return the upload_length as size.
```
Acconut, line 571: :100: Nice!

lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,131 @@
...
+
+                return resolve();
+            });
+        });
+    }
+
+    _newChunk() {
+        return new Promise((resolve, reject) => {
+            this.currentChunkPath = path.join(this.directory, `${this.filenameTemplate}-${this.part}`);
+            fs.open(this.currentChunkPath, 'w', (err, fd) => {
```
Acconut, line 115: I am wondering why you are using file descriptors here directly, instead of using streams? Is there some benefit? Easier usage?

lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,131 @@
...
+            return Promise.resolve();
+        }
+
+        return new Promise((resolve, reject) => {
+            fs.close(this.fileDescriptor, (err) => {
+                if (err) {
+                    return reject(err);
+                }
+
+                this.emit('chunkFinished', this.currentChunkPath);
```
Acconut, line 99: Will `emit` wait until all event listeners have been invoked? I am wondering if it is possible that somehow `chunkStarted` from `_newChunk` is invoked before chunkFinished?

lib/models/StreamSplitter.js
```diff
@@ -0,0 +1,131 @@
...
+
+                // The chunk fits in the max chunk size
+                return this._writeChunk(chunk)
+                    .then(() => callback())
+                    .catch(callback);
+            })
+            .catch(callback);
+    }
+
+    _final(callback) {
```
Acconut, line 64: Is _final a method that also gets called automatically by the parent stream.Writable class, just like _write?

Acconut (commented):
Awesome work, Merlijn! The code is looking amazing, I mostly just have questions to ensure everything is working as intended.

lib/stores/S3Store.js
Murderlon: I currently don't care about formatting or stylistic linting rules because after this PR I will start again in TS with a completely different Eslint setup and formatting with Prettier.

lib/stores/S3Store.js
Murderlon: `ReadableStream` is an implementation of the [WHATWG Streams Standard](https://streams.spec.whatwg.org/), currently stage 1 (experimental). It's mostly made to be isomorphic between Node.js and the browser, but it's significantly slower than the older streams API.

Also `pipeline` is a series of streams, wheres `ReadableStream` is a _single_ read stream.

lib/stores/S3Store.js
Murderlon: Yes, but we only know the file path after chunk started

lib/stores/S3Store.js
Murderlon: the extended `_write` classes which is automatically called in the `pipeline` will close the stream. The underlying class, `Stream.Writeable`, also emits events whenever an error occurs. We also close the stream then:

```js
        this.on('error', this._finishChunk.bind(this));
```

lib/stores/S3Store.js
Murderlon: Good question, this is from @mitjap's implementation and I'm not sure why we need it tbh

lib/stores/S3Store.js
Murderlon: linter is not happy indeed. But that will change in the rewrite

lib/models/StreamSplitter.js
Murderlon: We do use streams, but we create a tmp file for the stream to write to.

lib/models/StreamSplitter.js
Murderlon: `chunkStarted` is emitted and _afterwards_ resolved. We won't end up in the next `then` (where `chunkFinished` happens) until that promise is resolved

lib/stores/S3Store.js
Acconut: Ok, maybe my question was ill-phrased. Let's try again: Why do you use `stream.pipeline` here and not `req.pipe(splitter).on('error', ...)`?

lib/stores/S3Store.js
Acconut: > the extended _write classes which is automatically called in the pipeline will close the stream

Ok, sounds good.

lib/stores/S3Store.js
Acconut: Ok, I don't think it is necessary but let's leave it for now

lib/models/StreamSplitter.js
Acconut: Yes, but we have you decided to use `fs.open` and `fs.write(fd, ...)` directly instead of using `fs.createWriteStream` and `steam.write`? Are these too high level?

Acconut (approved):
Nice work! One question is still open, but that should not be a stopper from merging this

lib/stores/S3Store.js
Murderlon: I see! `pipeline` is an improved version added later in Node 10 while `pipe` is a lot older` It does a better job at cleaning up if one of the streams failed.

Acconut: Ah good to know, thank you!

lib/models/StreamSplitter.js
Murderlon: Right, the pattern of `open` and `write` has been used more often in this project. Both would do the job. I suppose we use it here to do some separation of concerns. `createWriteStream` takes care of opening and closing the stream as well, but we do some side-effects when that happens (so when we call `open` for instance).

tomgreco:
What exactly is the point of this? For what is the use case? I had assumed this was to support chunk sizes of less than 5mb but it is not...

Murderlon:
I would write it up again but this comment already does so: https://github.com/tus/tus-node-server/issues/137#issuecomment-863029907

tomgreco:
@Murderlon thank you for providing me that explanation. Are there any plans in the road map to support uploading less than 5MB chunks/partSizes to the server and potentially storing them in a buffer until at least 5MB of chunks are received and then uploading that to S3?

Murderlon:
@tomgreco definitely! It's on my todo list. I think I will be doing a sprint towards 1.0.0 around July which will be a complete rewrite in TypeScript, with proxy support, improved event emits, ability to configure configstore, and chunks <5mb.

I may consider to tackle chunks before the rewrite but I'm not sure yet

PR #254, opened 2022-06-14
index.d.ts: fix http import to work without default exports set

erikjwaxx:
Without `allowSyntheticDefaultExports` set to `true` in `tsconfig.json`, `import http from "http"` will cause typechecking to fail.

AFAIK explicitly changing this to `import * as http from "http"` as with the `fs` import above it should be safe regardless of the synthetic default export setting.


Murderlon:
Makes sense, thanks

PR #255, opened 2022-06-14
index.d.ts: explicitly type Server.get callback parameter signature

erikjwaxx:
Updates the type of the `Server.get` `callback` parameter to explicitly set its `...args` parameter as `any[]`.

Otherwise, with `noImplicitAny` set in tsconfig, tsc will complain about the implicit `any` type.


Murderlon (approved):

Murderlon:
Thanks!

PR #256, opened 2022-06-15
Sanitize file URLs

OzTK:
The current behaviour is creating a bad file url when the `path` is set to the root:
```js
    server.datastore = new tus.FileStore({
      path: '/'
    })
```
results in a URL of the form `http://localhost//12334456`

This sanitizes the constructed url when posting the download so that it results in a valid URL for the created file. Trimming was added as well.


Murderlon:
Thanks for the PR! But this will be superseded by #322

Issue #260, opened 2022-06-21
termination extension


mitjap:
FileStore does not properly report to client that 'termination' extension is supported as required by the protocol.

> If this extension is supported by the Server, it MUST be announced by adding termination to the Tus-Extension header.

mitjap:
Closed by #261

PR #261, opened 2022-06-21
announce to client that FileStore supports termination extension

mitjap:
This PR solves issue reported in #260.

### DIFF
--- a/lib/stores/DataStore.js
+++ b/lib/stores/DataStore.js
@@ -42,6 +42,10 @@ class DataStore extends EventEmitter {
         this._extensions = extensions_array;
     }
 
+    hasExtension(extension) {
+        return this._extensions.indexOf(extension) !== -1;
+    }
+
     /**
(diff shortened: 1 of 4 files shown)


mitjap:
I would further add checks to install DELETE request handlers ONLY if `termination` extension is supported. Current implementation falls back to `DataStores`'s implementation of delete which always succeeds. 

I guess this would be for another PR?

```
if (store.hasExtension('creation') || store.hasExtension('concatenation')) {
    this.handlers.POST = new PostHandler(store);
}

if (store.hasExtension('termination')) {
    this.handlers.DELETE = new DeleteHandler(store);
}
```

Murderlon (approved):

Murderlon:
> I would further add checks to install DELETE request handlers ONLY if `termination` extension is supported. Current implementation falls back to `DataStores`'s implementation of delete which always succeeds.
> 
> I guess this would be for another PR?

That would be nice as well.

PR #262, opened 2022-06-22
Reject partial upload if concatenation extension is not supported

mitjap:
Server returns error 400 and obvious error message if client wants to use concatenation extension when it is not supported.

### DIFF
--- a/lib/handlers/PostHandler.js
+++ b/lib/handlers/PostHandler.js
@@ -16,7 +16,7 @@ class PostHandler extends BaseHandler {
      */
     send(req, res) {
         if ('upload-concat' in req.headers && !this.store.hasExtension('concatentation')) {
-            return Promise.resolve(super.send(res, 400, {}, 'Concatenation extension is not supported'));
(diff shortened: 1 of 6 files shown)


lib/handlers/PostHandler.js
```diff
@@ -15,6 +15,10 @@ class PostHandler extends BaseHandler {
      * @return {function}
      */
     send(req, res) {
+        if ('upload-concat' in req.headers && !this.store.hasExtension('concatentation')) {
+            return Promise.resolve(super.send(res, 400, {}, 'Concatenation extension is not supported'));
```
Murderlon, line 19: I think the extra hint on how to fix it is helpful
```suggestion
            return Promise.resolve(super.send(res, 400, {}, 'Concatenation extension is not (yet) supported. Disable parallel uploads in the tus client. '));
```

Murderlon (commented):
I still think `501` (not implemented) is more appropriate than `400` (bad request). When I get a bad request, I'm more inclined to think the issue is on the client, and thus file an issue with tus-js-client fo instance. The client may have it perfectly implemented it but the server isn't ready for it yet. We should 'blame' the server with a server error response.

orshih6:
hello, parallel request not working.

PR #269, opened 2022-07-12
Correctly throw error in S3 store and check for file ID in GCSDataStore

mitjap:
Errors in S3 store are not throw correctly. `PostHandler` can properly handle errors that have `status_code` properties. If error object is encapsulated in `Error` class then such error can not be properly handled.

This PR also fixes some tests that did not work at all.

### DIFF
--- a/test/Test-Stores.shared.js
+++ b/test/Test-Stores.shared.js
@@ -1,4 +1,5 @@
 const assert = require('assert')
+const should = require('should');
 const sinon = require('sinon')
 const fs = require('fs')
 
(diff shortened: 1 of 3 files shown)


test/Test-Stores.shared.js
```diff
@@ -1,4 +1,5 @@
 const assert = require('assert')
+const should = require('should');
```
mitjap, line 2: For some reason my test are failing without this `require`.

Murderlon (approved):

Murderlon:
Thank you!

PR #270, opened 2022-07-18
S3Store: fix broken upload with deferred length or without metadata

mitjap:
Fix broken upload with deferred length or without metadata using S3 datastore. This fix is already in #186. This PR will allow us to merge this sooner.


Murderlon (approved):
Thanks! I'll take a look at the bigger PR soon.

Murderlon:
Tests are green :)

PR #271, opened 2022-07-18
Try to fix GH Actions for PRs from forks again

Murderlon:
Unfortunately, #267 and #268 didn't fix it. Another attempt.


Murderlon:
Some tests fail but that should only happen because I touched the worlflow file (because it's going to try to run the version present on the base branch). Hoping this works in forks now.

PR #280, opened 2022-08-02
Fix actions checkout

Murderlon:


Murderlon:
Any change to the CI script will fail initially on the PR but will be reflected correctly in a new PR.

PR #282, opened 2022-08-10
Allow `credentials` instead of key/secret for the S3 store

phillip-causing:


index.d.ts
```diff
@@ -22,11 +22,12 @@ declare interface GCStoreOptions extends DataStoreOptions {
 }
 
 declare interface S3StoreOptions extends DataStoreOptions {
-    accessKeyId: string;
-    secretAccessKey: string;
+    accessKeyId?: string;
+    secretAccessKey?: string;
+    credentials?: Object;
```
Murderlon, line 27: This is not strict enough, we can now omit  `accessKeyId`, `secretAccessKey`, _and_ `credentials`.

1. Change types to allow only `credentials` _or_ `accessKeyId` and `secretAccessKey`.
   1. Here is an [TS playground example](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wBYxmOAzKCEOAIixxIG4AoUSWOAbzgFEA3AUwDsYWRhsHKHAC++QsRIdOPVFWrUAJhwDGAG2RQOcYDyF5kK7QBFkMZAGUY0DgHkwMYBC7pG1OB7hgzGAFxxUGChdAHMaTzguZH4uEIAxAFcuFUdnAH5-AAotAEd-cjAAOgBJZKJQgFkOVFRkEI4ASjgAXgA+AKDQ8M8tDUdOABkIFTMnLgy4ACMICDUOZC4aEQVldU1tXRh9Q20LAGYrGwAhNDsHMfQOAA8trkV0U3NDrXtUlyZ3T0mElQBrDhg-kCwVi3Q8WhCYwmwK6nw8MHAxmAUAAClo8MArtDOqC4V5NDALMAAF4cfxcBIgSZCJZKVQaLQ6PRQAxGOD7Z5nN6oADqAgwAGkOABPACCdwsqi08DcEUMRhqQuFxUUQJxYTxqClANFKgVqCVapBGoiKi0yh4wGQalQEy40hp1GWqwZG2ZrN2B2sL3Ozl5-IAwubuI5ra48fLqgaRSq7Q6oGCAtqYLr9Uq45wE3izRwLaGbf5bJMAFaqGCUOAAekrcBAyH+cBgGGA6AW7L2jeFYG0ToUMC7ns5rwuLTgmQ53q5Fz5TaDuZDVptcAAZO3OSctcO-U0AD5jic2LcuGeCkXixSSnPwVcHrQbqfbhTXejwF3rOBrGpwR6WSdwa63PcrAyLw-CCMIsqeCofpBD83qZBAvouP4P5DkhqANIm9TwAB3CoBcmQNP4CzComWo4TceEEbhLgXAA+poUDIMKxFcMKmHZloZgcNkHB5HAKIEouAA8BQlGUMQhFUNR1BwrREQJEgthwwkka0iYAO7BFsmR4hEuT5NgRSlNBknSbU9QADR6Z4GJzHRwCKNixrWREEQQHgeDkXalLUlAeIKWi5RaqpbHqXi2G2J55GZHZHAOaqHTGoFSkhWptLUM+0CvvS76fugt7aDRQGoX+kEeNBLiwSk0AIeh-iFUeGFYQCUVeQCsXAPZjlGqEllwBp-J0d4sC2v40yzPMXApcFKnpb21DVh+0BaCk1D2hpa6TpkzDfH8AL+AA5Id-UjYSJJknAACMp0+EdJ0fsGlphv4jAiCIDTrRwm2FTtUw-P8gJwMdt2wESpL+Dd+JNvd-WRoqMaJSDSZXqmUaGsDh0fQoS1ac4IRfT9Xo2H9e2A7D+JgxdkO3TDmPYxtW0k7tAMHZjoPnRD120347OPfOz0FkwIhw3q6OIxTWqo2LCMsfTn1LVwEDwBiVy5nAwoAoTTNaKTrNA8jZ3g5dUPeHTyM5nmi6oP4STKBi9qKKLaYS3zUvSmjsv3Qz306zxLP7QbD1G9T3PQ7zFtPfmNvC874vKoldscA7ub9e7Ooy9GctJynigfUAA) that is almost perfect. Feel free to improve it to catch the missing cases. 
2. `import type` the actual credentials type from the aws lib.

Murderlon (commented):
Thanks the PR!

Couple things, and we should also add documentation to let users know this is supported.

index.d.ts
Murderlon: If we don't export the `S3StoreOptions` type, then it's probably better to do method overloading ([example](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wBYxmOAzKCEOAIixxIG4AoUSWOAbzgFEA3AUwDsYWRhsHKHAC++QsRIdOPVFWrUOAD3rxgPIXmQBjDnAAiyGMgDKMaBwDyYGMAhd0janBdwwRjAC44qGFHUA5jSucFzI-FwBAGIArlzatvYA-N4AFFAcAI7e5GAAdACS8USBALIcqKjIARwAlHAAvAB8Pn6Bwa4ZADZGwJwAMhDavcneAEYQEF0cyFw0IgrKqnDqMJo6eiYAzGYW1okOcMprXAAm6IbGuxn7dodOIWMx2gDWHDDevv6RHS4ZAXcUq1vkFnK4YOB9MAoAAFDJ4YBKIFfdpglzuWAmYAALw43i4MRAYyE80WKmg8FOHG0PQycBpaAuRlM5jpx2451YMl4-EEwgerm09i+z1ZqQgNjuqG8lxZe0lwtqvzgNXg7IcUtStW8swAnsrUO8jkoTqhNeqzcKAPrIKBQZC6nVcXVKtH0jJGDjpLLeGG22zILoAHlyhWKEQC5Uq1Q4TW1cDhJUNQb1TWVAHd-GtUm6QhlsphsPkikKI1GqjUADS51wI6ZW4CnZFtSLVkIhCB4PCGmBAglEoRu+OJ-jJ1PK1WWLs91J1jgN06fFsBYcSYBj51p6gLRTkhhUhl0hmVODba56dWc2XnphuoUOPyi6DihUObxn1lWV-oABkTBQ2i6JUADSHC6gUi7AoElY+NSGQwAAgoBFSoKBjpQZEoiuiE94igkz4Sgc0qnjsn63MKcB-sw2gZFSPDAIGxEQGMABW1LwCI2GuJO07vLOwD1o2S4gjB6YCBgVoYjAqBAhMUwzFwq5JhwKabqS1AAPQafS0AZAk1BcBw6YkeeqTME8rzvN4ADk1kwVJWK4t4ACM9keDZdnuhwdEBl0xGMCInEGUZJmfmZcAWW8HxwLZbmYjieJwK5bjuTFnk6EBqFgRBHkwYaNHvEhmVoR5QWadpmb2AEwXGR+FjhZFVlpXFMCOYlyXuDAXhpUFhm1aR9Xmc8UW5Sl8VOUlbldaNBU+QxfneAFMEZShaE5c1sEFYhyEgWBpW1DVoWDRFw1NbFY2tQlLlTd152zdwvnEXEVIIoZpzLTtWXgZBz0cK93l5XBhWfSVcC-f9py9SFdUZA1p3RedDlXZNKXTRt930Yxi0iB9xXZZB535fBRWrXtYNnH96jeVD-WmUNlkI55SMTR1qV3bRD3zU9FMQ5xQA)) 🤔

README.md
```diff
@@ -44,6 +46,22 @@ $ npm install tus-node-server
     });
     ```
 
+    using [credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html#constructor-property) (the example below uses [ECSCredentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS.html#ECSCredentials-property))
```
Murderlon, line 49: the first link refers to `AWS.Credentials`, which we don't use, so is that relevant?

The second link states that the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` or the `AWS_CONTAINER_CREDENTIALS_FULL_URI` environment variable needs to be present for this to work? If so, we should document that.

Lastly, we should explain what `ECSCredentials` does. I'm also wondering if `credentials` can take other classes then  `ECSCredentials`? Do you know?

Murderlon (commented):
Thanks for the update. I would like the test this functionality too in some test before merging I think. I'm not sure yet though what it takes to write a test for this.

README.md
phillip-causing: So if your application is gonna run inside ECS container the ECS will inject this env variable on your container and the `AWS.ECSCredentials` class will do the thing to resolve it into Key/Secret.
Yeah `AWS.ECSCredentials` is just one of the subclass of `AWS.Credentials`

index.d.ts
```diff
@@ -79,7 +79,8 @@ export declare class GCSDataStore extends DataStore {
  * file store in AWS S3
  */
 export declare class S3Store extends DataStore {
-    constructor(options: S3StoreOptions);
+    constructor(options: S3StoreOptions & { accessKeyId: string, secretAccessKey: string });
+    constructor(options: S3StoreOptions & { credentials: ECSCredentials });
```
Murderlon, line 83: If we allow all `credential` types, then the specific type of `ECSCredentials` is not what we want here. I think it might be the `RemoteCredentials` type?

README.md
```diff
@@ -44,6 +46,41 @@ $ npm install tus-node-server
...
+    - AWS.EnvironmentCredentials
+    - AWS.FileSystemCredentials
+    - AWS.ProcessCredentials
+    - AWS.RemoteCredentials
+    - AWS.SAMLCredentials
+    - AWS.SharedIniFileCredentials
+    - AWS.SsoCredentials
+    - AWS.TemporaryCredentials
+    - AWS.TokenFileWebIdentityCredentials
+    - AWS.WebIdentityCredentials
```
Murderlon, line 66:
```suggestion
    using [credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html#constructor-property) to fetch credentials inside a AWS container, such as an ECS container, which will inject the required environment variables. The `credentials` config is directly passed into the AWS SDK so you can refer to the AWS docs for the supported values for `credentials`.

File: index.d.ts (hunk 3 above)
Comment (phillip-causing) [reply index.d.ts]: Yeah you're right
Or perhaps use the `AWS.Credentials` itself

File: index.d.ts [hunk 5]
```diff
@@ -1,6 +1,8 @@
 import { EventEmitter } from "events";
 import * as fs from "fs";
 import * as http from "http";
+import { Credentials } from 'aws-sdk'
```
Murderlon, line 4:
```suggestion
import { RemoteCredentials } from 'aws-sdk'
```

index.d.ts
```diff
@@ -79,7 +79,8 @@ export declare class GCSDataStore extends DataStore {
  * file store in AWS S3
  */
 export declare class S3Store extends DataStore {
-    constructor(options: S3StoreOptions);
+    constructor(options: S3StoreOptions & { accessKeyId: string, secretAccessKey: string });
+    constructor(options: S3StoreOptions & { credentials: Credentials });
```
Murderlon, line 83:
```suggestion
    constructor(options: S3StoreOptions & { credentials: RemoteCredentials });
```

README.md
```diff
@@ -44,6 +46,26 @@ $ npm install tus-node-server
...
+        bucket: 'bucket-name',
+        credentials: new AWS.ECSCredentials({
+            httpOptions: { timeout: 5000 },
+            maxRetries: 10,
+        }),
+        region: 'eu-west-1',
+        partSize: 8 * 1024 * 1024, // each uploaded part will have ~8MB,
+        tmpDirPrefix: 'tus-s3-store',
+    });
+    ```
```
Murderlon, line 67:
```suggestion
using [credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html#constructor-property) to fetch credentials inside a AWS container, such as an ECS container, which will inject the required environment variables. The `credentials` config is directly passed into the AWS SDK so you can refer to the AWS docs for the supported values for `credentials`.

For example, with `ECSCredentials`:
    
    ```js
    server.datastore = new tus.S3Store({
        path: '/files',
        bucket: 'bucket-name',
        credentials: new AWS.ECSCredentials({
            httpOptions: { timeout: 5000 },
            maxRetries: 10,
        }),
        region: 'eu-west-1',
        partSize: 8 * 1024 * 1024, // each uploaded part will have ~8MB,
        tmpDirPrefix: 'tus-s3-store',
    });
    ```
```

Murderlon (commented):
`Credentials` is not correct (it doesn't have `httpOptions` and such) and something went wrong with your copy paste of my docs suggestion I think because it has unfinished sentences:

> The `credentials` config is directly passed into the AWS SDK so.)

Lastly, I don't think it's possible to write a test for this because `tus-node-server` would need to be inside a container?

Murderlon:
Thanks for all the changes! I'm holding this off for #309, which is now being worked on the `1.x` branch.

Murderlon (approved):

PR #283, opened 2022-08-10
EVENT_UPLOAD_COMPLETE checked in upload-length in last chunk

anuraagvaidya:
When upload-length-deferred is used in the case of streams, currently the FileStore does not emit EVENT_UPLOAD_COMPLETE event because it does not check the upload-length field of the request header. That's what this commit fixes.

### DIFF
--- a/lib/stores/FileStore.js
+++ b/lib/stores/FileStore.js
@@ -159,7 +159,7 @@ class FileStore extends DataStore {
                 log(`[FileStore] write: File is now ${offset} bytes`);
 
                 const config = this.configstore.get(file_id);
-                if (config && parseInt(config.upload_length, 10) === offset) {
+                if (config && parseInt(config.upload_length || req.headers['upload-length'], 10) === offset) {
                     this.emit(EVENTS.EVENT_UPLOAD_COMPLETE, { file: config });
                 }
                 return resolve(offset);


lib/stores/FileStore.js
```diff
@@ -159,7 +159,7 @@ class FileStore extends DataStore {
                 log(`[FileStore] write: File is now ${offset} bytes`);
 
                 const config = this.configstore.get(file_id);
-                if (config && parseInt(config.upload_length, 10) === offset) {
+                if (config && parseInt(config.upload_length || req.headers['upload-length'], 10) === offset) {
```
Murderlon, line 162: Thanks for the PR, but I'm not sure if this is the right place. In #186 we are splitting the responsibility of stores, they just handle store things and headers and request related things happen in request handlers.

Perhaps @mitjap has some thoughts on this here

anuraagvaidya: I agree. If you're splitting concerns then this can go in a separate callback, but since the event was not firing upon a live stream ending, this was the only immediate place I could find. I have applied it as a patch. Feel free to move it somewhere else.

mitjap:
This fix is relevant for both current master and #186 for all datastores and not only because of missing event.
But I'd do it differently. When we receive first chunk with `upload-length` header data in config store should be updated, then current check would work as expected. This might be easier to implement after #186 in merged.

Murderlon:
I can take a look at it after we merged #186 👍

Murderlon:
Superseded by #297

PR #284, opened 2022-08-11
Fix too small parts with S3 datastore

mitjap:
This PR fixes issue when uploading parts that are smaller than 5MB by ignoring them. Current implementation would upload such parts which would cause the upload to fail at the end when multipart upload is finished and S3 would reject such request.

This PR "fixes" two scenarios. 
First is when uploading chucks smaller than 5MB. Now all chunks are rejected and upload will never complete. Not sure if this is any better then error at the end :smile: 
Second is when chunks are larger than 5MB. Such upload are split to temporary files as specified in S3Store option `options.partSize` with default value of 8MB. Uploaded chunk is split into 8MB parts and a leftover part, this last part can be too small (<5MB) to be accepted by S3. This PR discards this part and signals client that it should be uploaded again. When upload is complete there will be no error.

First scenario still needs to be solved.

### DIFF
--- a/lib/models/StreamSplitter.js
+++ b/lib/models/StreamSplitter.js
@@ -96,7 +96,7 @@ class FileStreamSplitter extends stream.Writable {
                     return reject(err);
                 }
 
-                this.emit('chunkFinished', this.currentChunkPath);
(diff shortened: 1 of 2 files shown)


Murderlon (approved):

Murderlon:
Merci

PR #296, opened 2022-09-26
Fix upload ID when using S3 store with Digital Ocean Spaces

manelio:
Digital Ocean Spaces (S3 compatible object storage) sends the upload id information in metadata as `upload-id` instead `upload_id`, so the `_getMetadata` method of `S3Store` returns an empty upload id.

This change is a simple patch that works as a fallback to `upload-id` if `upload_id` is not found.


manelio:
Done!

Murderlon (approved):

Murderlon:
Than you!

PR #297, opened 2022-09-26
Fix creation-defer-length extension for all stores

Murderlon:
Supersedes #283

### DIFF
--- a/lib/handlers/PatchHandler.js
+++ b/lib/handlers/PatchHandler.js
@@ -41,7 +41,8 @@ class PatchHandler extends BaseHandler {
         }
 
         const new_offset = await this.store.write(req, file_id, offset);
-        if (`${new_offset}` === stats.upload_length) {
+        const upload_length = stats.upload_length || req.headers['upload-length'];
+        if (parseInt(new_offset, 10) === parseInt(upload_length, 10)) {
             this.emit(EVENTS.EVENT_UPLOAD_COMPLETE, { file: new File(file_id, stats.upload_length, stats.upload_defer_length, stats.upload_metadata) });
         }
 
--- a/lib/handlers/PatchHandler.js
+++ b/lib/handlers/PatchHandler.js
@@ -24,6 +24,7 @@ class PatchHandler extends BaseHandler {
         if (offset === undefined) {
             throw ERRORS.MISSING_OFFSET;
         }
+        offset = parseInt(offset, 10);
 
         // The request MUST include a Content-Type header
         const content_type = req.headers['content-type'];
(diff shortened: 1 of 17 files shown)


lib/handlers/PatchHandler.js
```diff
@@ -41,7 +41,8 @@ class PatchHandler extends BaseHandler {
         }
 
         const new_offset = await this.store.write(req, file_id, offset);
-        if (`${new_offset}` === stats.upload_length) {
+        const upload_length = stats.upload_length || req.headers['upload-length'];
+        if (parseInt(new_offset, 10) === parseInt(upload_length, 10)) {
```
mitjap, line 45: This approach would not work properly for S3 which need this information to split parts to chunks and only last chunk can be <5MB. If S3 can not determine that this is in fact last part it will be discarded and write will not finish.

Upload length should be written to datastore's configstore before any write is performed. Once this header is present it MUST be present and unchanged (and validated by the server) in all subsequent request.

From TUS protocol:
> If the length was deferred using Upload-Defer-Length: 1, the Client MUST set the Upload-Length header in the next PATCH request, once the length is known. Once set the length MUST NOT be changed. As long as the length of the upload is not known, the Server MUST set Upload-Defer-Length: 1 in all responses to HEAD requests

mitjap:
I'd rename `getOffset` to `getMetadata` and create new function `setMetadata` or `updateMetadata` to modify existing metadata. In `index.d.ts` we need to modify `IFile` and add property `size: number`. With this ready this issue should be easy to fix :)

mitjap:
This PR is almost complete. All that is needed is to implement `declareUploadLength` on `S3` and `GCS` datastores and enable tests via `shared.shouldDeclareUploadLendth()`.

Any idea why status 412 is sent when upload-length does not match? I think this error code is not appropriate and 400 would be much better.
```
INVALID_LENGTH: {
        status_code: 412,
        body: 'Upload-Length or Upload-Defer-Length header required\n',
    },
```

@Murderlon sorry for taking over this PR without consulting you. :blush:

Murderlon:
> Any idea why status 412 is sent when upload-length does not match? I think this error code is not appropriate and 400 would be much better.

Yes we can change that, tus also does this:
https://github.com/tus/tusd/blob/6ca6ef69a21c60dc0d00bfc55c4a34fc6398cf1b/pkg/handler/unrouted_handler.go#L60:2

Murderlon:
I'll take a look at the S3 and GCS ones

Murderlon:
Thanks for improving this too. I added it to S3 and GCS too. Two things:

- We have a race condition in CI when running the tests in parallel across node versions. It seems the CRUD operations against GCS don't allow these parallel changes and we get rate limited. Not sure what to do about that, I'm considering testing against one node version for now.
- Was quite confused for an hour that GCS has the very unobivous behavior when setting some meta data value to `undefined`, it actually ignores that and holds the previous value..

Regardless, this PR should be ready

mitjap:
I have no further comments. I think this can be released, 0.8.0 would be appropriate as there are some breaking changes.

PR #298, opened 2022-09-27
Do not run GitHub Actions in parallel

Murderlon:
There seems to be a race condition which causes GCS tests to fail. Disabling parallel jobs may fix this.


Murderlon:
Need to merge this to actually test it..

PR #299, opened 2022-09-30
Fix issue where there is no error checking bucket exists in GCSDataStore

angelsk:
When the bucket exists then `error` is `null` which causes this code to fail with an error  `TypeError: Cannot read properties of null (reading 'code')`.

We are using `"@google-cloud/storage": "^6.5.2",` but there's no harm in checking if a variable is not null before trying to access a property.

This is with the latest `"tus-node-server": "^0.8.0",` release

### DIFF
--- a/lib/stores/GCSDataStore.js
+++ b/lib/stores/GCSDataStore.js
@@ -46,7 +46,7 @@ class GCSDataStore extends DataStore {
         const bucket = this.gcs.bucket(this.bucket_name);
         bucket.exists((error, exists) => {
             // ignore insufficient access error, assume bucket exists
-            if (error.code === 403) {
+            if (error && error.code === 403) {
                 return;
             }
 


angelsk:
Sorry, I don't know why the test runs are failing - seems to be a timeout issue. Don't know how quick releases are made - this seems to be a fairly well maintained product :). But I am going to sadly have to downgrade the version until this can be patched coz my docker won't run.

Hope this helps

mitjap:
Thank you for this fix. This was recently changed. I wonder how that went past our tests. I recommend we also take a look at tests to see why this didn't trigger any errors.

Yeah, CI is grumpy sometimes :)

angelsk:
> Thank you for this fix. This was recently changed. I wonder how that went past our tests. I recommend we also take a look at tests to see why this didn't trigger any errors.
> 
> Yeah, CI is grumpy sometimes :)

It happens in the constructor - is that mocked somehow in the tests?

mitjap:
I guess your docker does not start because you have installed `uncaughtException` listener (https://nodejs.org/api/process.html#event-uncaughtexception) that terminates your application.

Thrown errors are unhandled and silently ignored. Intended behavior can not work as bucket existence can not be asynchronously checked in constructor. I recommend that we print warning message with logger and continue. Test should handle `uncaughtException`s and fail.

angelsk:
Are you able to handle the test part of this? I'm a PHP dev by trade, learning JS and no idea how to approach this beyond I found the error :)

mitjap:
Sure. We'll handle the rest.

mitjap:
@angelsk Version 0.8.1 is released. It should fix your issues.

angelsk:
> @angelsk Version 0.8.1 is released. It should fix your issues.

Thank you!

PR #302, opened 2022-09-30
Revert "Bump tus-js-client from 2.3.1 to 3.0.0"

mitjap:
Reverts tus/tus-node-server#289

This upgrade broke `npm run demo`. Problem is that `tus-js-client` 3.0.0 does not install demo folder anymore which is served by our `demo` command.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "should": "^13.2.3",
         "sinon": "^13.0.1",
         "supertest": "^6.2.2",
-        "tus-js-client": "^3.0.0"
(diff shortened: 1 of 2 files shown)


Murderlon:
Wouldn't the proper solution be to either `.gitkeep` or create the folder in the script?

Murderlon:
Oh I misunderstood. We can revert for the time being but I think we have to rewrite the demo to properly use `tus-js-client`. Or perhaps with `@uppy/core`/`@uppy/tus`

Murderlon (approved):

PR #303, opened 2022-10-01
fix validation of upload-length header in patch handler

mitjap:
I made a mistake in #297 thinking that either `Upload-Length` or `Upload-Defer-Length` headers MUST be present. This PR fixes this.

We don't need to validate `Upload-Defer-Length` as it is not expected to be present. `Upload-Length` can ONLY be present if extension is supported, and upload length is not yet known.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -64,6 +64,10 @@ const ERRORS = {
         status_code: 501,
         body: 'Concatenation extension is not (yet) supported. Disable parallel uploads in the tus client.\n',
     },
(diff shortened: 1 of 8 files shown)


Murderlon (approved):

Murderlon (approved):

Murderlon:
Test is still failing:

```
   1 failing

  1) PostHandler
       send()
         events
           must not fire the EVENT_UPLOAD_COMPLETE event when upload-length is defered:
     Error: done() invoked with non-Error: {"status_code":501,"body":"creation-defer-length extension is not (yet) supported.\n"}
  
```

PR #311, opened 2022-10-09
Clarify example for Fastify

piranna:
Some linters (or personal preference) remove `async` from functions if there are no `await` or return a `Promise` object. This, with some bad errors management in Fastify, has lead me to two full week of head banging. With the sync signature, it gets more explicit that the body parsing always succeed by doing nothing.

Note: `null` is needed due to how Typescript types are defined, for regular Javascript just calling `done()` should works.


Murderlon (approved):
Thank you!

piranna:
You are welcome :-)

PR #316, opened 2022-10-13
Added type of parameter on namingFunction

phillip-causing:


Murderlon:
Thanks, however I'm rewriting everything in TS (first PR: #308, second one with proper types coming). So this is superseded :)

Issue #300, opened 2022-09-30
GCS and S3 bucket inexistence handling


mitjap:
I'm opening a discussion that started in #299  about bucket existence checking. GCSDataStore does this asynchronously in constructor which by definition can not be async and any proper handling would cause race condition. If there is any error with this check or bucket does not exists all thrown errors are unhandled and usually causes application to exit with error code (crash).  
https://github.com/tus/tus-node-server/blob/9eeb26c40042f1f95ccba71f0ff51f792ca0c6ac/lib/stores/GCSDataStore.js#L47

S3Store handles bucket existence when uploading a file and in case of an error signals this in response with status code 500.
https://github.com/tus/tus-node-server/blob/9eeb26c40042f1f95ccba71f0ff51f792ca0c6ac/lib/stores/S3Store.js#L88

Murderlon:
I think it makes most sense to align with the S3 store here. During the initial POST it seems reasonable to find out at that time the bucket doesn't exist.

mitjap:
`POST` is optional extension and might not be used. 

I recommend testing at startup and save result as a promise which is always evaluated on every request. If initial test failed, all subsequest requests will fail. Another option is to just ignore this check (we can still print a warning to logs) and let it error on later requests. For example GCS reports that IAM does not have sufficient access if bucket does not exists when creating new object (on upload creation).

Murderlon:
I mean the initial POST _to S3_ not to the tus server, which is how it currently works for the S3 store, and apply that to GCS too. The way I see it, is that setting up a bucket is a prerequisite and we don't have to go too fancy.

Murderlon:
AFAIK this check doesn't explicitly exist in tusd, it just fails the creating request to S3 for instance. Do we really want to separate these failures? If you can't recover from it, and creating a bucket is a prerequisite, why do we need it?

Acconut:
Yes, tusd does not check the existence of buckets. It is not because I am opposed to the idea, but that I have not seen it to be big problem for users, so it is not a high priority.

If you add such a check, I would recommend to just check one upon startup and not very every upload. That seems just like a wasted request for every upload to me. Instead of putting it inside the constructor, stores could have a `checkSetup` method, which users can call to validate the setup (e.g. the bucket configuration).

PR #317, opened 2022-10-17
Implement TS types properly

Murderlon:
- Implement proper types everywhere as strictly as possible
- Tweak eslint config
- Create `tsconfig.dev.json` for the `test:build` command which also type checks test files
- Fix the getter/setter for `this.extensions`, they need to be the same type (previously array and string)
- In `GetHandler`, use a `Map` for `this.paths`, which is more appropriate.

### Questions
- I'm not sure how to type tests that use a handler. Handlers have `req` and `res`, which are objects with a lot of things. In tests we use simple things like `const req = { url: '/foo' }` and then the types fail. Same goes for `res`, the types say we can't pass a `method` to the constructor, yet it seems to work?

### For another PR
- Rewrite to use `node:fs/promises` and to use the promise version of the aws-s3 lib, which you can do with `async`/`await`, instead of `.promise().then()`.


Murderlon:
Working on fixing eslint errors

lib/stores/FileStore.ts
```diff
@@ -1,45 +1,55 @@
...
   }
 
   /**
    *  Ensure the directory exists.
    */
-  _checkOrCreateDirectory() {
+  private checkOrCreateDirectory() {
     fs.mkdir(this.directory, MASK, (error) => {
       if (error && error.code !== IGNORED_MKDIR_ERROR) {
         throw error
```
Acconut, line 55: I agree with the comment in the constructor, that throwing an error which cannot be handled at all it not the best approach :)

lib/stores/FileStore.ts
```diff
@@ -49,19 +59,16 @@ class FileStore extends DataStore {
...
+  create(file: File): Promise<File> {
     return new Promise((resolve, reject) => {
-      return fs.open(path.join(this.directory, file.id), 'w', async (err, fd) => {
+      fs.open(path.join(this.directory, file.id), 'w', (err, fd) => {
         if (err) {
           log('[FileStore] create: Error', err)
           return reject(err)
         }
 
-        await this.configstore.set(file.id, file)
```
Acconut, line 64: Why did you remove the await? Is the configstore synchronous?

lib/stores/FileStore.ts
```diff
@@ -180,14 +166,14 @@ class FileStore extends DataStore {
           return reject(ERRORS.FILE_NOT_FOUND)
         }
 
-        config.size = stats.size
-        return resolve(config)
+        file.size = stats.size
```
Acconut, line 169: Is it not problematic to modify the returned object from configstore.get? Considering the MemoryConfigStore, which only stores the objects in a big map, one request can modify the entry without calling configstore.set. Maybe we should do a deep clone in configstore.get?

lib/stores/GCSDataStore.ts
```diff
@@ -141,11 +121,11 @@ class GCSDataStore extends DataStore {
           return
         }
 
-        let bytes_received = (data as any).size
-        readable.on('data', (buffer: any) => {
+        let bytes_received = data.size as number
```
Acconut, line 124: Why do we need `as number` here? Shouldn´t we be able to control the return type from getOffset to always contain this numeric property?

lib/stores/GCSDataStore.ts
```diff
@@ -175,22 +155,15 @@ class GCSDataStore extends DataStore {
...
     return new Promise((resolve, reject) => {
       if (!file_id) {
         reject(ERRORS.FILE_NOT_FOUND)
         return
       }
 
-      const file = this.bucket.file(file_id)
-      file.getMetadata((error: any, metadata: any) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this.bucket.file(file_id).getMetadata((error: any, metadata: any) => {
```
Acconut, line 166: Are you still working on this type or is this intended to be left as-is?

Acconut (commented):
Amazing work, @Murderlon!

lib/handlers/BaseHandler.ts
```diff
@@ -13,37 +17,28 @@ class BaseHandler extends EventEmitter {
...
   }
 
-  generateUrl(req: any, file_id: any) {
-    return this.options.relativeLocation
-      ? `${req.baseUrl || ''}${this.options.path}/${file_id}`
-      : `//${req.headers.host}${req.baseUrl || ''}${this.options.path}/${file_id}`
+  generateUrl(req: http.IncomingMessage, file_id: string) {
+    if (this.options.relativeLocation) {
+      // TODO
+      // @ts-expect-error baseUrl doesn't exist? Should this be `.url`?
```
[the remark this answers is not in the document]
Murderlon: I know, still need to figure it out. They do exist in express.

lib/stores/FileStore.ts
Murderlon: Yes, we're handling this in a separate issue (#300) and PR.

lib/stores/FileStore.ts
Murderlon: It never was async indeed

lib/stores/GCSDataStore.ts
Murderlon: `size` technically can be undefined, but it's a bit weird with TS to say that in a certain function it can't be, while in other it can. Hence the cast in places it must be defined.

lib/stores/GCSDataStore.ts
Murderlon: `error` and `metadata` are both `any` in the underlying types, but TS complains so I make it explicit too.

lib/handlers/PatchHandler.ts
```diff
@@ -61,8 +62,9 @@ class PatchHandler extends BaseHandler {
       file.upload_length = upload_length
     }
 
+    // TODO: req should be a stream?
     const new_offset = await this.store.write(req, file_id, offset)
```
Murderlon, line 66: @Acconut @mitjap in the category "how has this ever worked", we have a type error here as we pass a `http.IncomingMessage` to `write` which excepts a `stream.Readable`. All our tests pass streams, not requests, so this is confusing me

lib/stores/FileStore.ts
Acconut: Ok, but do we want it to be synchronous in the future as well? Maybe in the future we want to have a file-based config store, which needs to be async. Do you not think this could be the case?

Murderlon: I personally prefer to not do premature optimizations. If we ever introduce it, it would be a non-breaking change to add an `await` and change the type signature.

Acconut (approved):
All of my feedback got addressed, so LGTM

PR #318, opened 2022-10-25
Rename `getOffset` to `getUpload`

Murderlon:
`getOffset` implies returning a number, the offset. Instead we return more information so `getUpload` seems better suited and aligned with `tusd`.

### Questions

Do we want to rename the keys of the return value, and the `File` model too? From:

```ts
  id: string
  upload_defer_length?: string
  upload_metadata?: string
  upload_length?: string
  size?: number
```

To:

```ts
  id: string
  isDeferred?: boolean
  metadata?: string
  length?: string
  size?: number
```


Murderlon:
Alright, I'll create a PR for that next!

PR #322, opened 2022-10-31
Respect forwarded headers

### LINKED ISSUE #59: Consider server and browsers with creation url

### LINKED ISSUE #134: Impossible to disable CORS
On my project, all CORS are managed by ESP.
And I found out that this package adds 'Access-Control-Allow-Origin' header.
As an ESP.
Duplicating the same 'Access-Control-Allow-Origin' header leads to error.
Is it a possible way to disable this header in the package?

### LINKED ISSUE #170: Unable to run behind a reverse proxy
I'm trying to run, without success, the tus server as an express middleware, behind a reverse proxy hosted in a URL other than "/".

On the server side:
```
const tusServer = new tus.Server();
tusServer.datastore = new tus.FileStore({
  path: "/files"
});
const uploadApp = express();
uploadApp.all("*", tusServer.handle.bind(tusServer));
app.use("/uploads", uploadApp);
```

When I start an upload on the browser, it correctly creates it, but notice how `/uploads` has a prefix path. The server responds with an incorrect `location`, and the following PATCH request don't work well:

image

Going through the closed issues, I saw a fix mentioning an undocumented property `relativeLocation`, so I tried to set that to true:

image

image

File creation still works and now it only returns the filename, but then the PATCH request fails, which although now includes the correct path prefix, I think it might be missing the "files/" part of it.

Am I doing anything wrong or missing something? thanks for the help.

### LINKED ISSUE #195: False uploadUrl
I'm using tus-js-client. 
**https://development.interskill.id/backendv2/** is my API base base URL.
I don't really know what happened. But I got false upload URL from tus-node server.
Screenshot from 2021-11-08 13-21-57

Murderlon:
Closes #256
Closes #134
Closes #59
Closes #170 
Closes #195

- Add new server option: `respectForwardedHeaders`.
- Respect `X-Forwarded-Host` and `X-forwarded-Proto` or `Forwarded` headers.
- Fix bug when `path` is set to root (`/`).

### DIFF
--- a/lib/handlers/BaseHandler.ts
+++ b/lib/handlers/BaseHandler.ts
@@ -30,12 +30,13 @@ export default class BaseHandler extends EventEmitter {
   generateUrl(req: http.IncomingMessage, id: string) {
     const forwarded = req.headers.Forwarded as string | undefined
     const path = this.options.path === '/' ? '' : this.options.path
+    // @ts-expect-error baseUrl type doesn't exist?
+    const baseUrl = req.baseUrl ?? ''
     let proto
     let host
 
     if (this.options.relativeLocation) {
-      // @ts-expect-error baseUrl type doesn't exist?
-      return `${req.baseUrl || ''}${path}/${id}`
+      return `${baseUrl}${path}/${id}`
     }
 
     if (this.options.respectForwardedHeaders) {
@@ -58,8 +59,7 @@ export default class BaseHandler extends EventEmitter {
     host ??= req.headers.host
     proto ??= 'http'
 
-    // @ts-expect-error baseUrl type doesn't exist?
-    return `${proto}://${host}${req.baseUrl ?? ''}${path}/${id}`
+    return `${proto}://${host}${baseUrl}${path}/${id}`
   }
 
   getFileIdFromRequest(req: http.IncomingMessage) {
--- a/lib/handlers/BaseHandler.ts
+++ b/lib/handlers/BaseHandler.ts
@@ -3,6 +3,9 @@ import EventEmitter from 'node:events'
 import type {DataStore, ServerOptions} from '../../types'
 import type http from 'node:http'
 
+const reForwardedHost = /host="?([^";]+)/
+const reForwardedProto = /proto=(https?)/
+
 export default class BaseHandler extends EventEmitter {
   options: ServerOptions
   store: DataStore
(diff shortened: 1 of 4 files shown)


mitjap:
This solution might solve some (most?) cases but not all. To solve all cases I think we need to expose `generateUrl` and `getFileIdFromRequest` and let the user do their specific magic. 

User can either use build-in functions and pass `relativeLocation` and `respectForwardedHeaders` options or provide own  `generateUrl` and `getFileIdFromRequest` functions.

Murderlon:
> This solution might solve some (most?) cases but not all. To solve all cases I think we need to expose `generateUrl` and `getFileIdFromRequest` and let the user do their specific magic.

Any cases you're thinking of already? AFAIK there is not going to be a different use case other then the way we get the ID from the request?

> User can either use build-in functions and pass `relativeLocation` and `respectForwardedHeaders` options or provide own `generateUrl` and `getFileIdFromRequest` functions.

While I'm not immediately convinced, should we do this and we really want to cover _all_ dynamic cases, then I think we can get rid of `relativeLocation` and `respectForwardedHeaders` and instead only expose `generateFileUrl`:

```js
generateFileUrl(req, id, next) {
  if (/* some custom condition */) {
    return '/something'
  }
  // next is the `generateUrl` function as it currently  is in this PR
  // you can call it or ignore it
  return next(req, id)
}
```

```js
// internally
generateFileUrl = options.generateFileUrl ?? generateFileUrl
```

Murderlon:
> This solution might solve some (most?) cases but not all. To solve all cases I think we need to expose `generateUrl` and `getFileIdFromRequest` and let the user do their specific magic.

Any cases you're thinking of already? AFAIK there is not going to be a different use case other then the way we get the ID from the request?

lib/handlers/BaseHandler.ts
```diff
@@ -24,15 +27,34 @@ export default class BaseHandler extends EventEmitter {
...
+        const proto = reForwardedProto.exec(forwarded)?.[1]
+        if (proto && host) {
+          return `${proto}://${host}${path}/${id}`
+        }
+      }
+
+      const forwardHost = req.headers['X-Forwarded-Host']
+      const forwardProto = (req.headers['X-Forwarded-Proto'] ?? 'https') as string
+
+      if (['http', 'https'].includes(forwardProto) && forwardHost) {
```
Acconut, line 46: Using this condition, X-Forwarded-Proto will only be respected if X-Forwarded-Host is also set. Could we make it so that each header can also be supplied individually without the other?

Acconut (commented):
Nice work!

PR #328, opened 2022-11-01
Make datastore a constructor option

### LINKED ISSUE #319: Pass datastore in the constructor instead of with a getter/setter
Currently we set a datastore like this:

```js
const tus = require('tus-node-server');

const server = new tus.Server({ path: '/files' });
server.datastore = new tus.FileStore({ directory: './files' });
```

But I don't see why it shouldn't be:

```js
const tus = require('tus-node-server');

const server = new tus.Server({ 
  path: '/files',
  datastore: new tus.FileStore({ directory: './files' })
});
```

This takes away the risk that you are trying to do anything without having set a datastore first, as we can throw an error the constructor already and make the types easier.

Murderlon:
Closes #319

### DIFF
--- a/demo/server.js
+++ b/demo/server.js
@@ -10,22 +10,13 @@ const GCSDataStore = require('../dist/index').GCSDataStore
 const S3Store = require('../dist/index').S3Store
 const EVENTS = require('../dist/index').EVENTS
 
-const options = {path: '/files'}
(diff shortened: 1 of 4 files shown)


Murderlon:
Merging this, hindsight reviews still welcome

PR #329, opened 2022-11-02
Rename File model to Upload

Murderlon:
- Rename `size` to `offset` and make the type non-optional
- Rename `upload_length` to `size`
- Create `sizeIsDeferred` getter
- Rename `upload_metadata` to `metadata`

### Questions
- Should we rename `File` to `Upload` since we recently renamed the method that returns it to `getUpload`?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,8 @@
     "@typescript-eslint/no-empty-function": "off",
     "unicorn/prevent-abbreviations": "off",
     "unicorn/no-null": "off",
-    "unicorn/filename-case": "off"
+    "unicorn/filename-case": "off",
+    "unicorn/explicit-length-check": "off"
   },
   "ignorePatterns": ["demo", "dist", "node_modules"]
 }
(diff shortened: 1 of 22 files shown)


mitjap:
I agree with proposed changes including rename to `Upload`. Given all this changes what is a difference between `configstore.get(id)` and `datastore.getUpload(id)`. Is `getUpload` just more up-to-date data?

Should we allow for each datastore to store some internal data to `File` and thus to `configstore`. This will allow us to use common configstore interface later on.

Murderlon:
> Given all this changes what is a difference between `configstore.get(id)` and `datastore.getUpload(id)`. Is `getUpload` just more up-to-date data?

AFAIK currently configstore only exists in filestore as an alternative to tusd's `.info` files. In the case of `getUpload`, we use `configstore` for the metadata but still reference the actual file on disk for the `offset`. S3 and GCS offer ways to store metadata. 

> Should we allow for each datastore to store some internal data to `File` and thus to `configstore`. This will allow us to use common configstore interface later on.

I'm not sure if we need config stores for all datastores? We could switch to `.info` files too at some point instead and remove configstores all together?

PR #335, opened 2022-11-17
S3Store: calculate optimal part size

Murderlon:
**Before**: uploading a 100 GB file with the default part size of 8 MB would fail the upload as it would be 12.5K parts, while the maximum allowed by S3 is 10K.

**After**: calculating the optimal part size which never exceeds the 10K limit.

Other changes:
 - Modernize `StreamSplitter`
 - Tiny refactors

### DIFF
--- a/lib/stores/S3Store.ts
+++ b/lib/stores/S3Store.ts
@@ -8,24 +8,18 @@ import aws from 'aws-sdk'
 import debug from 'debug'
 
 import DataStore from './DataStore'
-import FileStreamSplitter from '../models/StreamSplitter'
+import StreamSplitter from '../models/StreamSplitter'
 import Upload from '../models/Upload'
 import {ERRORS, TUS_RESUMABLE} from '../constants'
 
 const log = debug('tus-node-server:stores:s3store')
 
-function calculateOffsetFromParts(parts?: aws.S3.Parts) {
+function calcOffsetFromParts(parts?: aws.S3.Parts) {
   // @ts-expect-error object is not possibly undefined
   return parts && parts.length > 0 ? parts.reduce((a, b) => a + b.Size, 0) : 0
 }
 
-type Options = {
-  accessKeyId: string
-  secretAccessKey: string
-  bucket: string
-  region?: string
-  partSize?: number
-}
+type Options = {bucket: string; partSize?: number} & aws.S3.Types.ClientConfiguration
 
 type MetadataValue = {file: Upload; upload_id: string}
 // Implementation (based on https://github.com/tus/tusd/blob/master/s3store/s3store.go)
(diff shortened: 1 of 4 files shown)


lib/stores/S3Store.ts
```diff
@@ -374,6 +356,48 @@ export default class S3Store extends DataStore {
...
+    let optimalPartSize: number
+
+    // When upload is smaller or equal to PreferredPartSize, we upload in just one part.
+    if (size <= this.preferredPartSize) {
+      optimalPartSize = size
+    }
+    // Does the upload fit in MaxMultipartParts parts or less with PreferredPartSize.
+    else if (size <= this.preferredPartSize * this.maxMultipartParts) {
+      optimalPartSize = this.preferredPartSize
+      // Prerequisite: Be aware, that the result of an integer division (x/y) is
```
Acconut, line 369: I think we have to double check if the divisions in the lines below are actually safe. I see that you copied the code from tusd. Go has proper integer division, where the result will always be an integer. In JavaScript, that is not the case.

lib/stores/S3Store.ts
```diff
@@ -374,6 +356,48 @@ export default class S3Store extends DataStore {
...
+      // While in almost all cases, we could skip the check for the remainder and
+      // just add +1 to every result, but there is one case, where doing that would
+      // doom our upload. When (MaxObjectSize == MaxPartSize * MaxMultipartParts),
+      // by adding +1, we would end up with an optimalPartSize > MaxPartSize.
+      // With the current S3 API specifications, we will not run into this problem,
+      // but these specs are subject to change, and there are other stores as well,
+      // which are implementing the S3 API (e.g. RIAK, Ceph RadosGW), but might
+      // have different settings.
+    } else if (size % this.maxMultipartParts === 0) {
+      optimalPartSize = size / this.maxMultipartParts
```
Acconut, line 388: I think this division is ok, because we check before is size is multiple of maxMultipartParts.

lib/stores/S3Store.ts
```diff
@@ -374,6 +356,48 @@ export default class S3Store extends DataStore {
...
+      // Having a remainder larger than 0 means, the float result would have
+      // digits after the comma (e.g. be something like 10.9). As a result, we can
+      // only squeeze our upload into MaxMultipartParts parts, if we rounded UP
+      // this division's result. That is what is happending here. We round up by
+      // adding +1, if the prior test for (remainder == 0) did not succeed.
+    } else {
+      optimalPartSize = size / this.maxMultipartParts + 1
```
Acconut, line 395: I am concerned about this division as it might not return an integer but a decimal value. Maybe just use

```suggestion
      optimalPartSize = Math.floor(size / this.maxMultipartParts) + 1
```

and update the comments above.

Acconut (changes requested):
I am a bit concerned about the calculation of the optimal chunk size, but that should be fixable quickly. Awesome work!

Acconut (approved):
Looks good now! Great idea with the use of `Math.ceil`!

Issue #309, opened 2022-10-04
Road to 1.0.0


Murderlon:
This is the mother issue for the road to 1.0.0, which marks the first semver stable release for tus-node-server.

Progress on this release happens on the `1.x` branch. From now on we should carefully land things in `master`, as merge conflicts between the same `.js` and `.ts` files are hard to resolve. Ideally we only work on `1.x` now, except for critical bugs.

## Planned

- [x] #158 + new modern tooling (@Murderlon)
- [x] Implement proxy support. Take [tusd proxy support](https://github.com/tus/tusd/blob/master/docs/faq.md#can-i-run-tusd-behind-a-reverse-proxy) as inspiration (@Murderlon)
  *  [https://github.com/tus/tus-node-server/issues/134](https://github.com/tus/tus-node-server/issues/134) 
  *  [https://github.com/tus/tus-node-server/issues/59](https://github.com/tus/tus-node-server/issues/59) 
  *  [https://github.com/tus/tus-node-server/issues/170](https://github.com/tus/tus-node-server/issues/170) 
  *  [https://github.com/tus/tus-node-server/issues/195](https://github.com/tus/tus-node-server/issues/195) 
- [x] Rename `getOffset` ([PR](https://github.com/tus/tus-node-server/pull/318))
- [x] Rename `File` model to `Upload` and rename keys.
- [x] https://github.com/tus/tus-node-server/issues/319
- [x] https://github.com/tus/tus-node-server/issues/336
- [x] https://github.com/tus/tus-node-server/issues/300
- [ ] Return `413 Request Entity Too Large` when the client sends more data than reported with `Upload-Length` ([PR](https://github.com/tus/tus-node-server/pull/310))
- [x] S3: support chunks under 5 MB.
- [x] S3: implement buffering to prevent hitting the 10K parts limit. If we send a 500 GB file, we need to make sure we set the part size to around 52 MB in order to prevent hitting the 10K parts limit
- [x] #61 
- [x] Rewrite the README (@Murderlon)
- [ ] Change `master` to `main`

## Up for debate

- Rethink `ConfigStore` ([issue](https://github.com/tus/tus-node-server/pull/206#issuecomment-1049682476))
- S3: add termination extension
- GCS: add termination extension

---

Opinions welcome!

Acconut:
This roadmap looks good to me! Do you need any specific help from my side as I am assigned as well?

Murderlon:
[...]

PR #341, opened 2022-11-28
S3Store: handle chunks under the minimum S3 part size

Murderlon:
Refs: #309 

- Handle chunks under the minimum S3 part size. It does so by uploading chunks under 5MB as a regular upload to S3 with key as the upload ID + `.part`. On the next call to `store.write()`, we check if there is a incomplete part and prepend the buffer from the returned request into chunk we were about to upload. We don't need a temporary file for this incomplete part on disk so we only need to send a delete request when we're done with it.
- Small optimizations to reduce network requests.
- New tests


lib/stores/S3Store.ts
```diff
@@ -1,10 +1,12 @@
 import os from 'node:os'
 import fs, {promises as fsProm} from 'node:fs'
 import stream from 'node:stream/promises'
+import {Readable} from 'node:stream'
```
aduh95, line 4:
```suggestion
import type {Readable} from 'node:stream'
```

lib/stores/S3Store.ts
```diff
@@ -234,40 +295,58 @@ export default class S3Store extends DataStore {
...
-          .then(() => {
-            // Skip chunk if it is not last and is smaller than 5MB
-            const is_last_chunk = size === offset
-            if (!is_last_chunk && chunkSize < 5 * 1024 * 1024) {
-              log(`[${metadata.file.id}] ignoring chuck smaller than 5MB`)
-              return
+        const deferred = pDefer<void>()
+
+        // Fork off in the background
+        ;(async () => {
```
aduh95, line 315: I don't think using a p-defer is useful here?
```suggestion
        const deferred = new Promise(async (resolve, reject) => {
```

lib/stores/S3Store.ts
```diff
@@ -123,9 +128,12 @@ export default class S3Store extends DataStore {
       request.Metadata.original_name = parsedMetadata.filename.encoded
     }
 
+    // TODO: rename `file` to `upload` to align with the codebase
```
Acconut, line 130: Would addressing this TODO break existing uploads? i.e. a new version of tus-node-server would not be able to read upload's that were created by older tus-node-server versions?

lib/stores/S3Store.ts
```diff
@@ -205,12 +214,21 @@ export default class S3Store extends DataStore {
...
 
-  /**
-   * Uploads a part/chunk to S3 from a temporary part file.
-   */
+  private partKey(id: string, isIncomplete = false) {
+    if (isIncomplete) {
+      id += '.part'
+    }
+
+    // TODO: introduce ObjectPrefixing for parts and incomplete parts.
```
Acconut, line 221: You want to do this in a future PR?

lib/stores/S3Store.ts
```diff
@@ -226,6 +244,49 @@ export default class S3Store extends DataStore {
...
+    try {
+      const data = await this.client
+        .getObject({
+          Bucket: this.bucket,
+          Key: id,
+        })
+        .promise()
+      return data.Body as Buffer
+    } catch {
+      return undefined
```
Acconut, line 270: We could have more fine-graned error handling here. Only if the object was not found, undefined would be returned. Otherwise the error is raised

lib/stores/S3Store.ts
```diff
@@ -234,40 +295,58 @@ export default class S3Store extends DataStore {
...
+
+        // Fork off in the background
+        ;(async () => {
+          try {
+            const partNumber = currentPartNumber++
+            const incompletePartId = this.partKey(metadata.file.id, true)
+            // If we received a chunk under the minimum part size in a previous iteration,
+            // we used a regular S3 upload to save it in the bucket. We try to get the incomplete part here.
+            const incompletePart = await this.getIncompletePart(incompletePartId)
+            const isFinalChunk = size === offset + partSize
```
Acconut, line 320: Maybe we need a check here if the upload length is not deferred.

lib/stores/S3Store.ts
Murderlon: That's correct and already the case actually on the `1.x` branch. But AFAIK that only matters for GET requests, which is not part of the core protocol. Should we do something about that?

lib/stores/S3Store.ts
Murderlon: Yes I thought so.

lib/stores/S3Store.ts
Acconut: If a fix is as simple as checking for the `file` or `upload` key in the metadata, I think we could implement it.

lib/stores/S3Store.ts
Murderlon: It's not possible to reach this code for deferred files, `declareUploadLength` would be called before we hit `store.write`:

https://github.com/tus/tus-node-server/blob/accb18b2f53d113ded2e176213f353ccc4de8f48/lib/handlers/PatchHandler.ts#L76-L80

lib/stores/S3Store.ts
Murderlon: So I was mistaken and by  "That's correct and already the case actually on the `1.x` branch" I meant that the keys were already renamed inside the object. So it's not as simple as that. Perhaps I should rename them back (such as `size` to `upload_length`).

lib/stores/S3Store.ts
Acconut: Ok, great!

PR #344, opened 2022-11-30
Overhaul the events and hook callback system

### LINKED ISSUE #336: Rethink events
We currently have four events: `EVENT_ENDPOINT_CREATED`, `EVENT_FILE_CREATED`, `EVENT_UPLOAD_COMPLETE`, `EVENT_FILE_DELETED`. They seem a bit inconsistent. I think it makes sense to align with other servers on this.

Both [tusd](https://github.com/tus/tusd/blob/master/docs/hooks.md#list-of-available-hooks) and [rustus](https://s3rius.github.io/rustus/hooks/) implement the following hooks/events:

- **pre-create**. This event will be triggered before an upload is created.
- **post-create**. This event will be triggered after an upload is created.
- **pre-finish**. This event will be triggered after an upload is fully finished but before a response has been returned to the client.
- **post-finish**. This event will be triggered after an upload is fully finished, meaning that all chunks have been transferred and saved in the storage
- **post-terminate**. This event will be triggered after an upload has been terminated, meaning that the upload has been totally stopped and all associating chunks have been fully removed from the storage.
- **post-receive**. This event will be triggered for every running upload to indicate its current progress. It will be emitted whenever the server has received more data from the client but at most every second.

I think these mostly make sense for the tus-node-server too, except for perhaps for the progress (post-receive) hook every second. Another thing to consider is that some of these, such as pre-finish, were intented as blocking hooks. But tus-node-server just uses an event emitter which is non-blocking. Thus this event may not make sense too.

Thoughts, @Acconut and @mitjap?

Murderlon:
Closes #336 
Refs: #309 

- Remove `EVENT_ENDPOINT_CREATED`, `EVENT_FILE_CREATED`, `EVENT_UPLOAD_COMPLETE`, `EVENT_FILE_DELETED`
- Create `POST_CREATE`, `POST_RECEIVE`, `POST_FINISH`, `POST_TERMINATE`.
- Create blocking callbacks `onUploadCreate` and `onUploadFinish`. Error returned will be send back to the client.

Currently the error thrown in the callbacks expects an object `{body: 'some message', status_code: 500}`, which is consistent with the rest of our errors and has a fallback here:

https://github.com/tus/tus-node-server/blob/accb18b2f53d113ded2e176213f353ccc4de8f48/lib/Server.ts#L176-L184

This will be properly documented. But as I'm completely rewriting the readme and splitting up the packages, I'll do that when I get to it.

### DIFF
--- a/lib/handlers/PatchHandler.ts
+++ b/lib/handlers/PatchHandler.ts
@@ -30,13 +30,13 @@ export default class PatchHandler extends BaseHandler {
       throw ERRORS.INVALID_CONTENT_TYPE
     }
 
-    const file = await this.store.getUpload(id)
+    const upload = await this.store.getUpload(id)
 
     // If a Client does attempt to resume an upload which has since
     // been removed by the Server, the Server SHOULD respond with the
     // with the 404 Not Found or 410 Gone status. The latter one SHOULD
     // be used if the Server is keeping track of expired uploads.
-    const creation = file.creation_date ? new Date(file.creation_date) : new Date()
+    const creation = upload.creation_date ? new Date(upload.creation_date) : new Date()
     const expiration = new Date(creation.getTime() + this.store.getExpiration())
     const now = new Date()
     if (
@@ -47,10 +47,10 @@ export default class PatchHandler extends BaseHandler {
       throw ERRORS.FILE_NO_LONGER_EXISTS
     }
 
-    if (file.offset !== offset) {
+    if (upload.offset !== offset) {
       // If the offsets do not match, the Server MUST respond with the 409 Conflict status without modifying the upload resource.
       log(
-        `[PatchHandler] send: Incorrect offset - ${offset} sent but file is ${file.offset}`
+        `[PatchHandler] send: Incorrect offset - ${offset} sent but file is ${upload.offset}`
       )
       throw ERRORS.INVALID_OFFSET
     }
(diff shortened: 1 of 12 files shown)


lib/handlers/PatchHandler.ts
```diff
@@ -30,13 +30,13 @@ export default class PatchHandler extends BaseHandler {
...
+    const upload = await this.store.getUpload(id)
 
     // If a Client does attempt to resume an upload which has since
     // been removed by the Server, the Server SHOULD respond with the
     // with the 404 Not Found or 410 Gone status. The latter one SHOULD
     // be used if the Server is keeping track of expired uploads.
-    const creation = file.creation_date ? new Date(file.creation_date) : new Date()
+    const creation = upload.creation_date ? new Date(upload.creation_date) : new Date()
     const expiration = new Date(creation.getTime() + this.store.getExpiration())
     const now = new Date()
```
aduh95, line 41: Creating `Date` object is quite expensive.
```suggestion
    const now = Date.now()
    const creation = upload.creation_date ? new Date(upload.creation_date).getTime() : now
    const expiration = creation + this.store.getExpiration()
```

Acconut:
This PR looks amazing, great work! Only two comments, that can also be addressed in the future. There is no event to obtain the upload progress, comparable to post-receive in tusd. Also, people often want to modify the response in the onUploadCreate and onUploadFinish callbacks (e.g. add custom headers for the client). But the callbacks in this PR only get passed the HTTP request, not the response. 

Both of these can be added in the future, but do you have intentions to address this with the framework laid by this PR?

Murderlon:
> But the callbacks in this PR only get passed the HTTP request, not the response.

Changed it pass it the response and also always return the response, modified or not. 

> There is no event to obtain the upload progress, comparable to post-receive in tusd

Added it to the patch handler

PR #370, opened 2022-12-29
@tus/s3-store: add `s3ClientConfig` option

Murderlon:
It's a bit odd to merge the tus store options with the S3 SDK client config options. The client config is also massive so it would pollute the types. Instead we now have a new options to pass to the S3 store: `s3ClientConfig`, which is directly passed to the SDK.

### DIFF
--- a/packages/s3-store/README.md
+++ b/packages/s3-store/README.md
@@ -67,7 +67,7 @@ but may increase it to not exceed the S3 10K parts limit.
 
 Options to pass to the AWS S3 SDK.
 Checkout the [`S3ClientConfig`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html)
-docs for the supported options. Configuring your preferred method of authentication and region is required.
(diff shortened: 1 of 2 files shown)


packages/s3-store/README.md
```diff
@@ -67,6 +63,12 @@ The preferred part size for parts send to S3. Can not be lower than 5MB or more
 The server calculates the optimal part size, which takes this size into account,
 but may increase it to not exceed the S3 10K parts limit.
 
+#### `options.s3ClientConfig`
+
+Options to pass to the AWS S3 SDK.
+Checkout the [`S3ClientConfig`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html)
+docs for the supported options. Configuring your preferred method of authentication and region is required.
```
Murderlon, line 70:
```suggestion
docs for the supported options. You need to at least set the `region`, `bucket` name, and your preferred method of authentication. 
```

Acconut (approved):

PR #376, opened 2023-01-02
@tus/server: save decoded metadata in Upload model

### LINKED ISSUE #104: Filename encoded
The tus js client automatically encodes the filename before adding it to the request, but the node server does not decode it. It would be nice to have automatic decoding instead of having to manually parse and decode the filename from `event.upload_metadata`.

```javascript
{ file:
   { id: '236420e05a53609ddbf5d3104ca55551',
     upload_length: '48462',
     upload_metadata: 'filename bmV3bWFuLmpwZw==' } }
```

Murderlon:
Closes #104

- Remove `parseMetadataString` from `s3-store`
- Reuse existing `Metadata` model to parse and stringify the `Upload-Metadata` header. This is already thoroughly tested in [`Metadata.test.ts`](https://github.com/tus/tus-node-server/blob/main/packages/server/test/Metadata.test.ts).
- Save object of decoded metadata in `Upload` model instead of a big string

This should make dealing with metadata in events and hooks a lot easier.


packages/server/src/handlers/BaseHandler.ts
```diff
@@ -1,12 +1,13 @@
...
 
 import type {ServerOptions} from '../types'
-import type {DataStore} from '../models'
+import type {DataStore, Upload} from '../models'
 import type http from 'node:http'
 
 const reExtractFileID = /([^/]+)\/?$/
 const reForwardedHost = /host="?([^";]+)/
 const reForwardedProto = /proto=(https?)/
+const own = Object.prototype.hasOwnProperty
```
aduh95, line 10: This looks unused. (aside: there should probably be a lint rule for that at some point)
```suggestion
```

Acconut, line 10: Is `own` actually used somewhere in this file?

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -1,12 +1,13 @@
 import EventEmitter from 'node:events'
 
 import type {ServerOptions} from '../types'
-import type {DataStore} from '../models'
+import type {DataStore, Upload} from '../models'
```
Acconut, line 4: Is `Upload` used in this file?

packages/server/test/HeadHandler.test.ts
```diff
@@ -67,11 +65,11 @@ describe('HeadHandler', () => {
     const file = new Upload({
       id: '1234',
       offset: 0,
-      metadata: 'filename d29ybGRfZG9taW5hdGlvbl9wbGFuLnBkZg==,is_confidential',
```
Acconut, line 70: Is there another test covering the encoding and decoding of metadata? Because this one does not cover this aspect after your changes.

packages/server/src/handlers/BaseHandler.ts
Murderlon: Updated the rule to `error`

packages/server/test/HeadHandler.test.ts
Murderlon: Yes, there is another test in this file:

```js
  it('should resolve with metadata', async () => {
    const file = new Upload({
      id: '1234',
      offset: 0,
      metadata: {is_confidential: undefined, foo: 'bar'},
    })
    fake_store.getUpload.resolves(file)
    await handler.send(req, res)
    assert.equal(res.getHeader('Upload-Metadata'), 'is_confidential,foo YmFy')
  })

```

Acconut (approved):

PR #381, opened 2023-01-24
@tus/server: fix header exception when using a load balancer

dmitriiz:
I met conditions when Server cannot handle request. The reasons is that RequestValidator missing validation functions for 'X-Forwarded-Host' and 'X-Forwarded-Proto' headers.


Murderlon (approved):
This class really needs to be refactored at some point but for now this looks good. Thanks!

PR #385, opened 2023-01-25
@tus/server: handle non-ASCII file IDs

mitjap:
Properly encode `fileId` when composing URL and then decode when parsing. This allows user to use any naming function that is compatible with storage used.


Murderlon (approved):
Makes sense, thanks!

PR #388, opened 2023-01-30
@tus/gcs-store: allow user to pass bucket instance to GCSStore

mitjap:
This will allow users to use already configured Bucket or use all possible options that GCS provides. 

In my case I have a service that is responsible for handling all the secrets and create a global `Storage` instance. This way I can reuse that object without repeating any code. 

As a side note, at the moment GCS does not emit File object on `POST_FINISH` event, so it is not possible to access the remote file without external `Bucket` instance.


Murderlon (commented):
Since we're in beta phase we can still publish breaking changes. Wouldn't you say it might be better to only settle on passing a bucket object so we can remove the `projectId` and `keyFilename` options? (I suppose we would have to make GCS a peer dependency then).

btw I'm on holiday till February 22 so I'm only checking up on a couple things but won't really be coding/releasing.

mitjap:
That is even better. I did not want to change too much, but just add new functionality in somewhat backward compatible way. Peer dependency is always a good option.

Current implementation accepts either a `Bucket`  object or Storage options and Bucket options that are directly passed to GCS library. When we finish with this PR I'd do separate PR for S3 to make similar interface.

Murderlon:
I understand. But if we want to do breaking changes, I think now is the time. I find the current combination of options a bit confusing as one options cancels out other options. 

> Enjoy your holidays.

Thank you :)

Murderlon:
@mitjap what do you think?

mitjap:
This PR should be complete now. But GCS tests need to be fixed before we merge this.

PR #396, opened 2023-02-16
add addUrlDataToLocation option to handle dynamic route

primemb:
Hi
in one of my projects with the express, I need to handle file upload with a dynamic path and I need that path later to verify and send them back to the user so I add small changes and a custom option that allow defining those data add to Location header or not


Murderlon:
Hi, generally is better to first create a issue to get a sense of the problem and the direction of the solution. I'm not convinced yet this makes sense for tus. We have hooks and events to perform mid and after side effects. In this case, couldn't you send along metadata with the request and and use the `onUploadFinish` hook to move the file elsewhere, if that's what you need?

Murderlon:
Closing due to inactivity. But we can continue the discussion here :)

PR #401, opened 2023-02-26
@tus/s3-store: add termination extension

naumf:


Murderlon (commented):
Thanks for the PR. We do still need tests for this.

You can enable a couple by adding this:

https://github.com/tus/tus-node-server/blob/aab29d89ef5a19f25e46bf417c876767e8dc814f/packages/file-store/test.ts#L93

But I think we should also add a test to that shared file:
- terminate during upload

packages/s3-store/index.ts
```diff
@@ -517,4 +522,31 @@ export class S3Store extends DataStore {
...
+          .abortMultipartUpload({
+            Bucket: this.bucket,
+            Key: id,
+            UploadId: upload_id,
+          })
+          .promise()
+      }
+    } catch (error) {
+      log('remove: No file found.', error)
+      throw ERRORS.FILE_NOT_FOUND
```
Murderlon, line 540: Wondering if it makes sense to look at the error, because not all errors mean the file is not found (meaning it's already deleted)?

We do something similar here: 

https://github.com/tus/tus-node-server/blob/95366fbac8ed692aaf1a54d9821fd4f7877fd54a/packages/s3-store/index.ts#L190-L199

Murderlon (commented):
This looks very good! Thank you!

Minor thing, we need to update the table here as well:

https://github.com/tus/tus-node-server/blob/95366fbac8ed692aaf1a54d9821fd4f7877fd54a/packages/s3-store/README.md?plain=1#L82

Murderlon (approved):
Great! Thanks for the big PR

PR #406, opened 2023-03-01
@tus/file-store: add support for async configstore

Murderlon:
Closes #384

PR #407, opened 2023-03-01
@tus/file-store: support async config stores & add Redis store

### LINKED ISSUE #384: FileStore: Async configstore to support databases (e.g. Redis) as configstore
At the moment the configstore implementation of the FileStore has to be synchronous. 
This makes it only possible to use an in-memory or file-based store.

To support databases like redis for storing the configuration the configstore needs to async.

@samal-rasmussen mentioned this already here: https://github.com/tus/tus-node-server/pull/206#issuecomment-1049682476

### LINKED ISSUE #440: Reconsider `configstore` package with our own implementation
[`configstore`](https://www.npmjs.com/package/configstore), according to the package itself:

> Easily load and persist config without having to think about where and how.
> The config is stored in a JSON file located in `$XDG_CONFIG_HOME` or `~/.config`.
Example: `~/.config/configstore/some-id.json`

The default location (which we don't change) is for config files, which really doesn't feel right for storing upload metadata. It's more expected to have them live next to your uploads. I think we could quite easily replace this with our own implementation and get rid of another dependency. Perhaps this should be done as a breaking change before we move out of beta.

Related: #407

mitjap:
Closes #384 
Closes #440 

I made some changes but it is still a draft. I'm still not sure how to handle dependencies between turbo packages, so this might not compile properly. 

I see there is also https://github.com/tus/tus-node-server/pull/406 that tries to implement same functionality.


Murderlon (commented):
I think this PR is the more elaborate approach. I personally wouldn't have included the Redis store and let everyone define their own custom stores, whatever those may be. But since the work is already done it doesn't hurt to ship it :)

Dependencies in turbopack aren't any different from how you use dependencies normally. However since the idea of config stores only exists in file store, I think it's better to move the modal and all stores over to that package. For tests in server, we only need `@tus/file-store` as a dev dependency.

mitjap:
> However since the idea of config stores only exists in file store

Are you sure about that? Personally I'd like to use GCS store with redis config store for reduced latency. 

For first iteration we can move everything to `@tus/file-store` and think about other uses in the future.

packages/file-store/test.ts
```diff
@@ -7,9 +7,10 @@ import path from 'node:path'
 import sinon from 'sinon'
 
 import {FileStore} from './'
-import {Upload, MemoryConfigstore} from '@tus/server'
+import {Upload} from '@tus/server'
```
Murderlon, line 10: I think we needed the `@tus/server` dev dependency because of this usage in tests.

Murderlon:
Are you still interested in working on this? It's one of the last things to move out of beta.

mitjap:
I would like to complete this task, but I'm not familiar with turbo, so it requires some extra effort on my part.

This PR introduced circular dependency between `@tus/file-store` and `@tus/server` which is not recommended and can cause some weird and unexpected issues.

packages/file-store/test.ts
mitjap: I belive `@tus/server` should be listed as a dependency because `FileStore` extends `DataStore` and uses other types from `@tus/server`. 
```
import {DataStore, Upload, ERRORS} from '@tus/server'
```

Sure, `@tus/server` is also used in tests but since it is declared as a dependency I don't think we need to repeat that in `devDependencies`.

packages/file-store/package.json
```diff
@@ -21,12 +21,10 @@
     "test": "mocha test.ts --exit --extension ts --require ts-node/register"
   },
   "dependencies": {
-    "configstore": "^5.0.1",
+    "@tus/server": "workspace:*",
     "debug": "^4.3.3"
   },
   "devDependencies": {
-    "@tus/server": "workspace:^",
```
mitjap, line 28: I see that this notation `workspace:^` is often used. But in documentation I only find `workspace:*`. Do you know what is a difference?

Murderlon: Here is an example of input and output:

```js
{
  "dependencies": {
    "star": "workspace:*",
    "caret": "workspace:^",
    "tilde": "workspace:~",
    "range": "workspace:^1.2.3",
    "path": "workspace:path/to/baz"
  }
}
```

```js
{
  "dependencies": {
    "star": "1.5.0",
    "caret": "^1.5.0",
    "tilde": "~1.5.0",
    "range": "^1.2.3",
    "path": "1.5.0"
  }
}
```

packages/file-store/test.ts
Murderlon: I think it makes more sense to have peer dependency for how people consume this package and a dev dependency for our tests. What do you think?

kompany-sde:
Hi @Murderlon @mitjap  - I am currently working on an application and this merge request would solve a critical issue I am currently facing due to the configstore not being async. Do you know roughly when this could be merged and published?

Murderlon:
@kompany-sde we can try to get this released next week!

packages/file-store/configstores/RedisConfigstore.ts
```diff
@@ -0,0 +1,37 @@
...
+import {Upload} from '@tus/server'
+import {Configstore} from './Types'
+
+/**
+ * Redis based configstore.
+ *
+ * @author Mitja Puzigaća <mitjap@gmail.com>
+ */
+export class RedisConfigstore implements Configstore {
+  constructor(private redis: RedisClientType, private prefix: string = '') {}
```
Murderlon, line 12:
```suggestion
  constructor(private redis: RedisClientType, private prefix: string = '') {
    this.redis = redis
    this.prefix = prefix
  }
```

packages/file-store/configstores/index.ts
```diff
@@ -0,0 +1,4 @@
+export {FileConfigstore} from './FileConfigstore'
+export {MemoryConfigstore} from './MemoryConfigstore'
+export {RedisConfigstore} from './RedisConfigstore'
+export {Configstore} from './Types'
```
Murderlon, line 4: nit: types should be lowercase `./types`

Murderlon (commented):
Code looks very good. Regarding dependencies, I still think we should have `@tus/server`as a peer dependency in every package, it can't be a dependency as people should determine their server version. Then to make tests work we add it as a dev dependency.

kompany-sde:
> @kompany-sde we can try to get this released next week!

@Murderlon - That would be awesome! Thanks! Great work you are doing with this lib btw.

packages/file-store/configstores/FileConfigstore.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(packageName?: string, options?: ConfigstoreOptions) {
+    if (!packageName) {
+      packageName = `${pkg.name}-${pkg.version}`
+    }
+    this.db = new Store(packageName, undefined, options)
+  }
+
+  async get(key: string): Promise<Upload | undefined> {
+    return this.db.get(this.encodeKey(key))
```
Acconut, line 18: I am a bit curious how the types work here. How can TypeScript verify that `this.db.get` returns an instance of `Upload`? I think the signature for the `Store#get` is `get(key: string): any;` (from https://www.npmjs.com/package/@types/configstore). So does this cast `any` into `Upload`?

packages/file-store/configstores/RedisConfigstore.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  async list(): Promise<Array<string>> {
+    return this.redis.keys(this.prefix + '*')
+  }
+
+  private serializeValue(value: Upload): string {
+    return JSON.stringify(value)
+  }
+
+  private deserializeValue(buffer: string | null): Upload | undefined {
```
Acconut, line 34: Similar question about types here. How is `JSON.parse(buffer)` turned into an `Upload`?

packages/file-store/index.ts
```diff
@@ -188,25 +182,29 @@ export class FileStore extends DataStore {
...
   }
 
   async deleteExpired(): Promise<number> {
     const now = new Date()
     const toDelete: Promise<void>[] = []
 
-    const uploadInfos = this.configstore.all
-    for (const file_id of Object.keys(uploadInfos)) {
+    if (!this.configstore.list) {
+      return 0
```
Acconut, line 201: Can we have a more explicit error here (or somewhere else) that expiring uploads is not supported if your config store does not support the list operation?

Or should we make the list operation just mandatory?

Acconut (commented):
This look good overall, but I am a bit confused about the types.

packages/file-store/configstores/FileConfigstore.ts
Murderlon: TypeScript is just a superset so we can't really rely on types like in other languages. It just trusts you that you're right.

mitjap: You could use [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) to make sure you get the types right in runtime.

packages/file-store/configstores/RedisConfigstore.ts
Murderlon: TypeScript thinks it's `any` but we just tell it to be `Upload`

packages/file-store/configstores/FileConfigstore.ts
```diff
@@ -0,0 +1,44 @@
+import fs from 'node:fs/promises'
+import {Upload} from '@tus/server'
+
+import {Configstore} from './Types'
+
+export class FileConfigstore implements Configstore {
```
Acconut, line 6: This class does not seem to be safe for concurrent access because the underlying file system operations are not atomic. For example, a concurrent call to `list` and `delete` can return incomplete data. It's probably not the end of the world, but worth mentioning here.

Acconut (commented):
One comment otherwise everything LGTM

packages/file-store/configstores/FileConfigstore.ts
Murderlon: Good point. `list` is only called in a CRON to clean expired uploads. That shouldn't happen often (or not at all if you don't use the extension). That leaves us with `delete` and `get`, for which I reckon that shouldn't really happen if there is only one client operating on the ID at hand.

We may be fine for now?

Acconut: In theory, most combinations of two methods here can lead to unexpected results, but the likelihood is probably low. So let's not get stuck here for now.

mitjap: This could be solved by queuing promises so that at any given time only one is running.

Murderlon: Good point. We could use something like [`p-queue`](https://github.com/sindresorhus/p-queue) with `concurrency: 1`.

Acconut: That might already be sufficient, yes.

Acconut (approved):

mitjap:
@Murderlon thank you for pushing this PR over the finish line :partying_face:

PR #430, opened 2023-05-11
Update incorrect comment

spencerbart:
https://tus.io/protocols/resumable-upload.html#:~:text=MUST%20be%20omitted.-,Upload%2DMetadata,-The%20Upload%2DMetadata


mitjap:
Good catch. I would reword this comment to
> If an upload contains additional metadata, responses to HEAD requests MUST include the Upload-Metadata header and its value as specified by the Client during the creation.

This is from specification (https://tus.io/protocols/resumable-upload.html).

mitjap (approved):

PR #441, opened 2023-06-07
@tus/s3-store: replace Underscores with Hyphens in Metadata Field Names

liknenS:
Hello,

I am proposing a change to the metadata field naming convention used in the AWS SDK S3 operations. Specifically, I recommend replacing underscores `_` with hyphens `-`.

### Motivation
In compliance with RFC 7230, HTTP headers should not contain underscores. However, I've noticed that the current implementation uses underscores in the metadata field names, which subsequently form part of the headers. This has the potential to lead to compatibility issues when interacting with services that strictly follow the standards. One such example is Nginx, which by default, ignores headers containing underscores.

### Problem Description
When using the AWS SDK S3 to interact with S3-like services(minio) behind a proxy server (like Nginx), the service may return an error stating that "There were headers present in the request which were not signed". This is due to the headers with underscores being ignored, leading to discrepancies between the expected and actual headers.

### Proposed Solution
To avoid such compatibility issues, I suggest modifying the metadata field naming convention to use hyphens instead of underscores. This aligns better with widely accepted HTTP header practices and prevents potential compatibility issues with proxy servers or other services.

This proposed change should be largely transparent to users but would significantly improve the robustness and compatibility of the service.

Looking forward to your feedback.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -26,7 +26,7 @@ type Options = {
   s3ClientConfig: aws.S3.Types.ClientConfiguration & {bucket: string}
 }
 
-type MetadataValue = {file: Upload; upload_id: string; tus_version: string}
+type MetadataValue = {file: Upload; 'upload-id': string; 'tus-version': string}
 // Implementation (based on https://github.com/tus/tusd/blob/master/s3store/s3store.go)
 //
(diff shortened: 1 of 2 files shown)


Murderlon (approved):
Makes sense! Thanks for the contribution :)

PR #443, opened 2023-06-12
replace type-fest by raw typescript

MichaelBitard:
We have some issue in our project with different versions of type-fest, and we noticed it is only used once in the whole project.

I rewrote the usage in typescript and got rid of the dependency.


MichaelBitard:
My bad :) it's fixed

Murderlon:
Releasing once #446 is merged :)

Murderlon:
Released https://github.com/tus/tus-node-server/releases/tag/%40tus%2Fserver%401.0.0-beta.6

Issue #444, opened 2023-06-14
[S3] Potential Race Condition - Corrupt files


fenos:
Hi,

I'm suspecting there is a race condition when running tus-node-server on an EBS or any network volumes.

When I pause an upload (canceling the request) the [pipeline stream](https://github.com/tus/tus-node-server/blob/main/packages/s3-store/index.ts#L289) throws an `aborted` error.

On the error handler, we issue an `fs.rm` call, which asynchronously removes the temporary chunk.
However, the `chunkFinished` callback on the `StreamSplitter`, is also invoked asynchronously while the chunk is being removed.

When running on an EBS or network volume the call to remove the chunk might take sometime and in certain laggy conditions I'm suspecting that the `chunkFinished` callback reaches the [uploadIncompletePart](https://github.com/tus/tus-node-server/blob/main/packages/s3-store/index.ts#L271) with only part of the original chunk.

Subsequentially, when resuming the corrupted part is appended and generates an invalid file.

I couldn't reproduce this on my local machine, because the fs.rm call is faster than [reading the chunk](https://github.com/tus/tus-node-server/blob/main/packages/s3-store/index.ts#L265) before uploading.

I found a few instances of large uploaded images that seemed to have been corrupted during the upload

I will try to dig deeper into this, but any thoughts would be appreciated

fenos:
What's the reason behind invoking the `chunkFinished` event on error in the SplitStream?

https://github.com/tus/tus-node-server/blob/main/packages/server/src/models/StreamSplitter.ts#L36

fenos:
I could reproduce the issue by mimicking latency during the [fs.rm](https://github.com/tus/tus-node-server/blob/main/packages/s3-store/index.ts#L293) call by doing this:

```
await new Promise((resolve, reject) => {
   setTimeout(() => {
         try {
             await fsProm.rm(pendingChunkFilepath)
         } catch(err) {
         } finally {
            resolve()
        }
    }, 500)
})
```

This [PR](https://github.com/tus/tus-node-server/pull/446) will prevent the `chunkFinished` event to be invoked in case of an error, potentially fixing this problem

PR #446, opened 2023-06-15
@tus/server: do not emit chunkFinished on error

fenos:
Related to https://github.com/tus/tus-node-server/issues/444 

If an error occurs during the write stream, we don't want to broadcast the chunkFinished event.
This should fix potential race conditions

### DIFF
--- a/packages/server/src/models/StreamSplitter.ts
+++ b/packages/server/src/models/StreamSplitter.ts
@@ -85,17 +85,20 @@ export class StreamSplitter extends stream.Writable {
     this.currentChunkSize += chunk.length
   }
 
-  async _finishChunk(): Promise<void> {
+  async _finishChunk(err?: Error): Promise<void> {
     if (this.fileHandle === null) {
       return
     }
 
     await this.fileHandle.close()
 
-    this.emit('chunkFinished', {
-      path: this.currentChunkPath,
-      size: this.currentChunkSize,
-    })
+    if (!err) {
+      this.emit('chunkFinished', {
+        path: this.currentChunkPath,
+        size: this.currentChunkSize,
+      })
+    }
+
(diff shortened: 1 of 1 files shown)


packages/server/src/models/StreamSplitter.ts
```diff
@@ -85,17 +85,20 @@ export class StreamSplitter extends stream.Writable {
...
       return
     }
 
     await this.fileHandle.close()
 
-    this.emit('chunkFinished', {
-      path: this.currentChunkPath,
-      size: this.currentChunkSize,
-    })
+    if (!err) {
```
Murderlon, line 95: Can we add an elaborate comment on why this is needed here?

Murderlon (commented):
Thanks a lot for looking into this!

fenos:
@Murderlon actually, I think a cleaner way to handle this is to have its own error handler function instead of mixing it into the `_finishChunk` function, for example:

```ts
  async _handleError() {
    if (this.fileHandle === null) {
      return
    }

    await this.fileHandle.close()
    this.currentChunkPath = null
    this.fileHandle = null
  }
```

It feels cleaner doing it this way, since we don't even need to advance the part to next chunk either.

Murderlon (approved):
Thank you!

Murderlon:
> @Murderlon actually, I think a cleaner way to handle this is to have its own error handler function instead of mixing it into the `_finishChunk` function, for example:

Probably a bit cleaner yes. Let's add it.

fenos:
@Murderlon Just sent the patch using its dedicated error handler function

Murderlon:
Released https://github.com/tus/tus-node-server/releases/tag/%40tus%2Fserver%401.0.0-beta.6

PR #454, opened 2023-07-12
@tus/s3-store: migrate to V3 SDK

### LINKED ISSUE #394: Use modularized AWS SDK v3 to reduce size
[Since version 3, AWS SDK is modularized](https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/). This allows to use only the subset of functionality required from the SDK and reduce the size of node_modules.
This is important in order to reduce the overall size of docker images: in my specific case, the main contributor to the final docker image size is the aws-sdk folder in node_modules (116MB), most of which comes from unneeded `.d.ts` and `.json` files related to many AWS services which are not used in tus-node-server.

Murderlon:
Closes #394

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -26,7 +26,11 @@ type Options = {
   s3ClientConfig: S3ClientConfig & {bucket: string}
 }
 
-type MetadataValue = {file: Upload; 'upload-id': string; 'tus-version': string}
+type MetadataValue = {
+  file: Upload
+  'upload-id': string
+  'tus-version': string
+}
 // Implementation (based on https://github.com/tus/tusd/blob/master/s3store/s3store.go)
 //
 // Once a new tus upload is initiated, multiple objects in S3 are created:
@@ -204,9 +208,13 @@ export class S3Store extends DataStore {
     })
   }
 
-  private async prependIncompletePart(path: string, part: Uint8Array): Promise<void> {
-    const file = await fsProm.readFile(path)
-    await fsProm.writeFile(path, Buffer.concat([file, Buffer.from(part)]))
+  private async prependIncompletePart(
+    newChunkPath: string,
+    previousIncompletePart: Uint8Array
+  ): Promise<void> {
+    const newChunk = await fsProm.readFile(newChunkPath)
+    const combined = Buffer.concat([previousIncompletePart, newChunk])
+    await fsProm.writeFile(newChunkPath, combined)
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/s3-store/index.ts
```diff
@@ -172,30 +169,26 @@ export class S3Store extends DataStore {
...
-      .promise()
+    const data = await this.client.putObject({
+      Bucket: this.bucket,
+      Key: id,
+      Body: readStream,
+    })
     return data.ETag as string
   }
 
-  private async getIncompletePart(id: string): Promise<Buffer | undefined> {
```
Acconut, line 185: Is there a reason for removing the return type here?

packages/s3-store/index.ts
```diff
@@ -205,18 +198,15 @@ export class S3Store extends DataStore {
...
+      Bucket: this.bucket,
+      Key: id,
+    })
   }
 
-  private async prependIncompletePart(path: string, buffer: Buffer): Promise<void> {
-    const part = await fsProm.readFile(path, 'utf8')
-    buffer.write(part, buffer.length - 1)
-    await fsProm.writeFile(path, buffer)
+  private async prependIncompletePart(path: string, part: Uint8Array): Promise<void> {
```
Acconut, line 207: This function confuses me a bit. The arguments' names are not easy to understand. I assume `path` points to the file with the data that we just received from the client and `part` is the incomplete part that we fetched from S3. If so, maybe the names could be improved or a comment could be added.

In addition, I think the order in `Buffer.concat` is wrong. Currently it appends the incomplete part, but we want to prepend it.

Acconut (commented):
Nice work! I only have two comments to add

packages/s3-store/index.ts
Murderlon: I'll try to clear it up. However, concat should be working as expected as the test passes:

https://github.com/tus/tus-node-server/blob/ca1bb7a3c87701dbfc70d16b867186daf89cd4fd/packages/s3-store/test.ts#L36-L42

Murderlon: Update: the test also passes if I flip the argument. Something is not right.

Murderlon: I used `assert()` instead of `assert.strictEqual()`...

Acconut: Oh, good one :D

AFAIK, the original logic before this PR and now after the PR are correct, right?

PR #470, opened 2023-08-24
@tus/s3-store: fix invalid character errors for "x-amz-meta-file” header

### LINKED ISSUE #458: @tus/s3-store: fix invalid character errors for "x-amz-meta-file” header
It seems we may naively stringifying a file here:

https://github.com/tus/tus-node-server/blob/0af377dfc4877b44c95238a3b46d96429cad96ca/packages/s3-store/index.ts#L429

Instead of stringifying the entire Upload, we may want to put all keys individually inside the metadata.

Murderlon:
Fixes #458 

This was a bit odd, instead of creating an `.info` file with the upload inside the file, we used to create an empty file and put everything on metadata, which the SDK sends alongs has query parameters. Sending an entire JSON string in a header is error-prone and caused issues. What's also odd, is that in `create` we put everything on metedata too but we _also_ call `saveMetadata`, which already does the job.

Now we actually create a file with contents, and only use metadata for `upload-id` and `tus-version`.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -98,9 +98,8 @@ export class S3Store extends DataStore {
     await this.client.putObject({
       Bucket: this.bucket,
       Key: `${upload.id}.info`,
-      Body: '',
+      Body: JSON.stringify(upload),
       Metadata: {
-        file: JSON.stringify(upload),
         'upload-id': uploadId,
         'tus-version': TUS_RESUMABLE,
       },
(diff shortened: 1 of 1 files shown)


Acconut (approved):
Great change! It is good to have this breaking change already in v1!

Murderlon:
> It is good to have this breaking change already in v1!

How breaking do you think this is? Biggest issue is with GET'ing files from S3, for which it will look on old files inside the file which is empty. GET is not officially supported by the protocol but can be implemented. Other than that, there should be no issues afaik. Important enough to keep supporting both even though we are technically allowed to make a breaking change in a beta release?

Acconut:
My question is whether an upload that is created by pre-PR tus-node-server can be resumed with the post-PR tus-node-server. I.e., does the "data format" change on S3 cause in-progress uploads to break when tus-node-server is upgraded. Is this the case?

Murderlon:
Before we used to check for the offset and such in the metadata, now that is in the `.info` file itself. So that would also be breaking if you deploy this and a user tries to resume one hour apart or so.

That is not great and backwards compatibility may make sense here. Although at some point we should get rid of it, so why not now which is allowed by semver. 

Maybe I should write a test to see if someone tries to resume the upload with the old format and the new code, that the server doesn't crash or get stuck for this upload.

PR #471, opened 2023-08-24
@tus/server: introduce onIncomingRequest middleware

### LINKED ISSUE #460: @tus/server: introduce middleware for access control
Access control can be done in many ways. We don't want to be opinionated on that so implementing a generic middleware function which is used before all requests seems to be the best approach.

Murderlon:
Closes #460

- [x] Add  `onIncomingRequest`
- [x] Docs
- [x] Tests

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -80,7 +80,7 @@ Default uses `crypto.randomBytes(16).toString('hex')`.
 `onUploadCreate` will be invoked before a new upload is created. (`(req, res, upload) => Promise<res>`).
 
 If the function returns the (modified) response, the upload will be created.
-If an error is thrown, the HTTP request will be aborted and the provided `body` and `status_code` (or their fallbacks) will be sent to the client.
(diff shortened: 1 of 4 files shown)


packages/server/README.md
```diff
@@ -93,6 +93,17 @@ If an error is thrown, the HTTP request will be aborted and the provided `body`
 
 This can be used to implement post-processing validation.
 
+#### `options.onIncomingRequest`
+
+`onIncomingRequest` is a middleware function invoked before all handlers (`(req, res, upload) => Promise<res>`)
```
Acconut, line 98: This documentation mentions an `upload` argument, which I do not see in the rest of the code. Do you plan on implementing this?

Murderlon: Ah that's a copy paste mistake. Because we run `onIncomingRequest` before any handler, no `Upload` is created yet so we can't send it along.

Acconut (approved):

PR #475, opened 2023-08-30
@tus/s3-store: fix min part size condition & offset race condition

### LINKED ISSUE #415: Last part of multi-upload is ignored
Hello, it seems the issue of "Last part of multi-upload is ignored" (https://github.com/tus/tusd/pull/462) is not yet fixed in tus/server and @tus/s3store, so now when I integrate @tus/server and @tus/s3store with Nodejs (specifically nextjs) I have the issue that last part of the uploads were ignored, thus the uploaded file is incomplete

Murderlon:
Fixes #415

- if the current part is exactly the minimum part size, then that should also be allowed, not minimum + 1
- fix race condition in which the stream splitter is faster than uploading to S3, resulting the offset not being updated in time before the next chunk is handled.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -257,8 +257,10 @@ export class S3Store extends DataStore {
             const readable = fs.createReadStream(path)
             readable.on('error', reject)
             if (partSize >= this.minPartSize || isFinalChunk) {
-              await this.uploadPart(metadata, readable, partNumber)
+              // The stream splitter may be faster than we are able to upload to S3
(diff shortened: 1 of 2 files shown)


Acconut (approved):
Good catch! Could we add a test for this to avoid a regression?

Murderlon:
It seems I missed another edge case. Taking another look next week and I'll make tests for both cases then.

Acconut (approved):
Oh, that's a nasty bug. Great work!

PR #490, opened 2023-10-05
@tus/file-store: fix list method of FileConfigstore

### LINKED ISSUE #488: Error when calling `server.cleanUpExpiredUploads()`
When `FileStore` is use as `datastore` and use the default `configstore` by not providing a `configstore`

```js
const datastore = new FileStore({
    directory: 'mydir',
    expirationPeriodInMilliseconds: 60000 * 60 * 24, // 24hrs,
});
```

Calling `server.cleanUpExpiredUploads()` will cause an error

Because in this line [packages/file-store/configstores/FileConfigstore.ts#L43](https://github.com/tus/tus-node-server/blob/main/packages/file-store/configstores/FileConfigstore.ts#L43)

```js
 async list(): Promise<Array<string>> {
    return this.queue.add(async () => {
      const files = await fs.readdir(this.directory, {withFileTypes: true})
      const promises = files
        .filter((file) => file.isFile() && file.name.endsWith('.json'))
        .map((file) => fs.readFile(path.resolve(file.path, file.name), 'utf8'))

      return Promise.all(promises)
    }) as Promise<string[]>
  }
```
`file.path` is `undefined` but I was able to fix it by replacing it to `this.directory`
I was able to read the file but there is another problem, in this line it is expecting just the `file_id/key` [packages/file-store/index.ts#L207](https://github.com/tus/tus-node-server/blob/main/packages/file-store/index.ts#L207) but what we have returned on the list above is an array of the content of the files

Murderlon:
Fixes #488

### DIFF
--- a/packages/file-store/configstores/FileConfigstore.ts
+++ b/packages/file-store/configstores/FileConfigstore.ts
@@ -38,6 +38,10 @@ export class FileConfigstore implements Configstore {
   async list(): Promise<Array<string>> {
     return this.queue.add(async () => {
       const files = await fs.readdir(this.directory)
+      // readdir returns files without file extension.
+      // This means we will get two files,
+      // the actual file and the JSON info file, per upload.
+      // We de-duplicate here to only return one ID per upload.
       return Array.from(new Set(files))
     })
   }
--- a/packages/file-store/configstores/FileConfigstore.ts
+++ b/packages/file-store/configstores/FileConfigstore.ts
@@ -38,10 +38,12 @@ export class FileConfigstore implements Configstore {
   async list(): Promise<Array<string>> {
     return this.queue.add(async () => {
       const files = await fs.readdir(this.directory)
-      // list is expected to return an array of IDs,
-      // which is the same as the file name.
-      // But the JSON info files are also in the directory so we filter those.
-      return files.filter((file) => !file.endsWith('.json'))
+      const sorted = files.sort((a, b) => a.localeCompare(b))
+      const name = (file: string) => path.basename(file, '.json')
+      // To only return tus file IDs we check if the file has a corresponding JSON info file
+      return sorted.filter(
+        (file, idx) => idx < sorted.length - 1 && name(file) === name(sorted[idx + 1])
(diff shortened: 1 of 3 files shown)


packages/file-store/configstores/FileConfigstore.ts
```diff
@@ -37,13 +37,13 @@ export class FileConfigstore implements Configstore {
...
-        .map((file) => fs.readFile(path.resolve(file.path, file.name), 'utf8'))
-
-      return Promise.all(promises)
-    }) as Promise<string[]>
+      const files = await fs.readdir(this.directory)
+      // readdir returns files without file extension.
+      // This means we will get two files,
+      // the actual file and the JSON info file, per upload.
+      // We de-duplicate here to only return one ID per upload.
+      return Array.from(new Set(files))
```
phillip-causing, line 45: I have tested this
`fs.readdir` still return files with extension
image

phillip-causing (changes requested):

packages/file-store/configstores/FileConfigstore.ts
Murderlon: Thanks for double checking! Not sure how I missed this.

packages/file-store/configstores/FileConfigstore.ts
```diff
@@ -37,13 +37,12 @@ export class FileConfigstore implements Configstore {
...
-        .filter((file) => file.isFile() && file.name.endsWith('.json'))
-        .map((file) => fs.readFile(path.resolve(file.path, file.name), 'utf8'))
-
-      return Promise.all(promises)
-    }) as Promise<string[]>
+      const files = await fs.readdir(this.directory)
+      // list is expected to return an array of IDs,
+      // which is the same as the file name.
+      // But the JSON info files are also in the directory so we filter those.
+      return files.filter((file) => !file.endsWith('.json'))
```
mitjap, line 44: I'd check the opposite and consider only files that have JSON info file present. This way you ignore any random files in this folder.

PR #492, opened 2023-10-16
@tus/s3-store Fix asynchronous processing of chunks

### LINKED ISSUE #415: Last part of multi-upload is ignored

mitjap:
This PR fixes #415.

Problem is that all logic regarding current part number, offset or determining whether this is final part must be synchronous. Failing to do so can cause parts to be reordered or not properly detected as final (more common case).

To achieve this all logic is moved out of asynchronous code and is executed synchronously in event callback.


packages/s3-store/index.ts
```diff
@@ -235,31 +238,36 @@ export class S3Store extends DataStore {
       .on('chunkFinished', ({path, size: partSize}) => {
         pendingChunkFilepath = null
 
+        const partNumber = currentPartNumber++
+        const chunkNumber = currentChunkNumber++
+
+        offset += partSize
+
+        const isFirstChunk = chunkNumber === 0
```
Murderlon, line 246: If you increment `currentChunkNumber` before this check, won't it always be `> 0`?

packages/s3-store/index.ts
```diff
@@ -235,31 +238,36 @@ export class S3Store extends DataStore {
...
-            const isFinalChunk = size === offset + partSize
-
-            if (incompletePart) {
-              // We found an incomplete part, prepend it to the chunk on disk we were about to upload,
-              // and delete the incomplete part from the bucket. This can be done in parallel.
-              await Promise.all([
-                this.prependIncompletePart(path, incompletePart),
-                this.deleteIncompletePart(incompletePartId),
-              ])
+            if (isFirstChunk) {
```
Murderlon, line 253: why do we only check for an incomplete part if it's the first chunk?

Murderlon (commented):
Thanks for looking into this

packages/s3-store/index.ts
mitjap: Only first chunk of each PATCH request can prepend incomplete part from previous request.

mitjap: I've used different terms to disambiguate. `chunk` is a part of a **current** request , while `part` is part of S3 (multipart) upload.

Murderlon (approved):

PR #493, opened 2023-10-16
@tus/s3-store: Fix offset calculation for uploads with incomplete part

mitjap:
This PR fixes incorrect `Upload-Offset` on HEAD request for uploads with incomplete part. 

First commit is the actual fix, second is non-mandatory refactor which makes `id` parameter having a consistent meaning for all functions in this class. Before this change `id` sometimes meant `<upload-id>`, but when working with incomplete parts it was expected to be `<upload-id>.part`.
Handling of suffix `.part` is moved into all functions that work directly with incomplete parts.

Note: this PR will conflict with #492 because of this second commit. I'll fix that when either is merged.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -486,7 +486,7 @@ export class S3Store extends DataStore {
       throw error
     }
 
-    const incompletePart = await this.getIncompletePart(this.partKey(id))
+    const incompletePart = await this.getIncompletePart(this.partKey(id, true))
 
     return new Upload({
       id,
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -170,7 +170,7 @@ export class S3Store extends DataStore {
   ): Promise<string> {
     const data = await this.client.putObject({
       Bucket: this.bucket,
-      Key: id,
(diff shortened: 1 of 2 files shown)


mitjap:
Tests are failing because there are no cleanups between runs. How do we handle that? Should I use random upload id?

Murderlon (commented):
Great find!

>  Tests are failing because there are no cleanups between runs. How do we handle that? Should I use random upload id?

That would be easiest for now yes. I have a lifecycle in S3 that cleans everything up after 24 hours so it wouldn't be filling up. Ideally we also do this for every test at some point, because then we can enable concurrency across Node versions in CI.

mitjap:
Many tests rely on the fact that when new upload is created existing upload with same `id` is overwritten. At the moment it is kind of undefined behaviour for `s3-store` (`<id>.part` and `<id>` files are not removed), but not for `filestore`. To avoid this UB we'd need to remove those two files, and the tests should work as expected.

Murderlon:
Sending a lot of delete requests to S3 will slow them the suite (and might incur higher costs?). If we always send unique IDs per test, we should be fine and let the lifecycle handle the rest?

PR #494, opened 2023-10-18
@tus/s3-store: Upload as multipart upload when incomplete part grows beyond minimal part size

mitjap:
This PR solves scenario when client chunk size was less then 5MB, which caused data to be written as incomplete part. But when that incomplete part grew beyond 5MB it was not "promoted" to part of multipart upload but kept growing. This meant that amount of data being downloaded and re-uploaded was growing until upload was complete. 

This behavior was only possible since #493, before this scenario wound not work.


Murderlon (approved):
👌

PR #495, opened 2023-10-19
@tus/s3-store: Don't download incomplete part on HEAD request

mitjap:
Currently on every HEAD request entire incomplete part is downloaded to memory. 

This PR makes HEAD request to S3 to determine incomplete part size.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -4,7 +4,7 @@ import stream from 'node:stream/promises'
 import type {Readable} from 'node:stream'
 import http from 'node:http'
 
-import AWS, {S3, S3ClientConfig} from '@aws-sdk/client-s3'
+import AWS, {NotFound, S3, S3ClientConfig} from '@aws-sdk/client-s3'
 import debug from 'debug'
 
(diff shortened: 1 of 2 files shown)


packages/s3-store/index.ts
```diff
@@ -197,6 +197,21 @@ export class S3Store extends DataStore {
...
+  private async getIncompletePartSize(id: string): Promise<number | undefined> {
+    try {
+      const data = await this.client.headObject({
+        Bucket: this.bucket,
+        Key: this.partKey(id, true),
+      })
+      return data.ContentLength
+    } catch (error) {
+      if (error instanceof NotFound) {
+        return undefined
```
Murderlon, line 209: nit: when there is no incomplete part when can also consider the size `0`. Especially if we want to always do addition on the return value.

```suggestion
        return 0
```

Murderlon (approved):
Nice improvement!

packages/s3-store/index.ts
mitjap: I was thinking about that. In this case we would not be able to tell if there is no incomplete part or if its size is `0`. For HEAD request that distinguishing is actually not needed. I'm good with both ways.

PR #496, opened 2023-10-19
@tus/s3-store: Stream incomplete part instead of reading to memory

mitjap:
This PR improves transferring and prepending of incomplete parts. This was all done in-memory, but now this is changed to use streams. 

Incomplete parts are usually smaller than 5MB but chunk size can be up to 5GB in size. Reading all this to memory for concatenation is not a good idea.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -1,10 +1,10 @@
 import os from 'node:os'
 import fs, {promises as fsProm} from 'node:fs'
-import stream from 'node:stream/promises'
+import stream, {promises as streamProm} from 'node:stream'
 import type {Readable} from 'node:stream'
 import http from 'node:http'
 
-import AWS, {NotFound, S3, S3ClientConfig} from '@aws-sdk/client-s3'
+import AWS, {NoSuchKey, NotFound, S3, S3ClientConfig} from '@aws-sdk/client-s3'
 import debug from 'debug'
 
 import {DataStore, StreamSplitter, Upload} from '@tus/server'
@@ -177,19 +177,15 @@ export class S3Store extends DataStore {
     return data.ETag as string
   }
 
-  private async getIncompletePart(id: string) {
+  private async getIncompletePart(id: string): Promise<Readable | undefined> {
     try {
       const data = await this.client.getObject({
         Bucket: this.bucket,
         Key: this.partKey(id, true),
       })
-      return data.Body?.transformToByteArray()
+      return data.Body as Readable
     } catch (error) {
-      if (
-        error.Code === 'NoSuchKey' ||
-        error.Code === 'NoSuchUpload' ||
-        error.Code === 'AccessDenied'
-      ) {
+      if (error instanceof NoSuchKey) {
         return undefined
       }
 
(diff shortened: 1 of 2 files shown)


packages/s3-store/index.ts
```diff
@@ -177,19 +177,15 @@ export class S3Store extends DataStore {
...
       const data = await this.client.getObject({
         Bucket: this.bucket,
         Key: this.partKey(id, true),
       })
-      return data.Body?.transformToByteArray()
+      return data.Body as Readable
     } catch (error) {
-      if (
-        error.Code === 'NoSuchKey' ||
-        error.Code === 'NoSuchUpload' ||
```
Murderlon, line 190: I think we also had `NoSuchUpload` for S3-compatible storages who return a different code

packages/s3-store/index.ts
```diff
@@ -221,11 +217,40 @@ export class S3Store extends DataStore {
...
+      await streamProm.pipeline(
+        previousIncompletePart,
+        byteCounterTransform,
+        fs.createWriteStream(tempPath)
+      )
+      // append to temporary file
+      await streamProm.pipeline(
+        fs.createReadStream(newChunkPath),
+        fs.createWriteStream(tempPath, {flags: 'a'})
+      )
```
Murderlon, line 243: Would this work too?

```suggestion
    await streamProm.pipeline(
      previousIncompletePart,
      byteCounterTransform,
      // Open file for appending. The file is created if it does not exist.
      fs.createWriteStream(tempPath, { flags: 'a' }),
      fs.createReadStream(newChunkPath),
      fs.createWriteStream(tempPath, { flags: 'a' })
    );
```

Murderlon (commented):
🙌 

> Incomplete parts are usually smaller than 5MB but chunk size can be up to 5GB in size

You mean theoretically people can configure it to be that high right? It's unlikely, but even at smaller sizes this approach is probably better.

packages/s3-store/index.ts
mitjap: No. Pipeline starts with readable stream, following zero or more transform streams and end with writable stream. What we need is to concatenate two streams.

I'm opening a file two times because it is less error prone to use `stream.pipeline(...)`. When pipeline is finished it closes write stream so it can not be reused for second pipeline. More efficient approach would be to manually `pipe` from both streams and manually handle errors and closing at the end.

mitjap:
> > Incomplete parts are usually smaller than 5MB but chunk size can be up to 5GB in size
>
> You mean theoretically people can configure it to be that high right? It's unlikely, but even at smaller sizes this approach is probably better.

This is the same distinction that we discussed in previous PR. Incomplete part is smaller than 5MB because it can not be uploaded as part. Chunk is part of a single PATCH request that will be uploaded as part of multipart upload. Size of this chunk is anywhere between 5MB to 5GB and is calculated automatically. There can be at most 10k chunks. That together limits the size of files on S3. 

Examples: 
Uploading a 100GiB file, a chunk size will be 10MiB.
Uploading a 1TiB file, a chunk size will be 100MiB.

https://github.com/tus/tus-node-server/blob/80dbbfa7fb74a1af4b5035b3d679133030479732/packages/s3-store/index.ts#L400-L417

When prepending the incomplete parts, both were read to memory.

packages/s3-store/index.ts
mitjap: I was looking for any previous discussion on this topic but could not find it. We can keep all types of error, but eventually I'd like to check which are actually required.

`NoSuchUpload` is returned for multipart upload operations where you provide `UploadId`.

Murderlon: I also can't find it anymore. So I'm okay with simplifying 👍

Murderlon:
Another great addition ✨

PR #498, opened 2023-10-19
@tus/s3-store: Method for reading the file is missing

andriykhc:
This PR addresses an issue where an uploaded URL file produces a not found response

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -469,12 +469,12 @@ export class S3Store extends DataStore {
    * Read file by upload id
    */
   read(id: string) {
-    const request: S3.GetObjectRequest = {
+    const object = this.client.getObject({
       Bucket: this.bucket,
       Key: id,
-    }
-    
-    return this.client.getObject(request).createReadStream()
+    })
+
+    return object.Body;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mitjap:
Your example is not valid GetObject command for latest @aws-sdk/client-s3 library.

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/getobjectcommand.html

This is what works for me. But this is asynchronous function which is not fully supported by the library.
```
async read(id: string) {
  const data = await this.client.getObject({
    Bucket: this.bucket,
    Key: id,
  })
  return data.Body as Readable;
}     
```

Another alternative is to synchronously return a Readable stream and pass data to it when object's stream is available.

```
read(id: string) {
  const pass = new stream.PassThrough();
  
  (async () => {
    try {
      const data = await this.client.getObject({
        Bucket: this.bucket,
        Key: id,
      })

      try {
        if (data.Body) {
          streamProm.pipeline(data.Body as Readable, pass);
        } else {
          pass.write(null);
        }
      } catch (err) {
        // ignore errors
      }
    } catch (err) {
      pass.destroy(err);
    }
  })()

  return pass;
}
```

andriykhc:
Sorry, I used this library on a project that still uses aws-sdk v2 and I forgot that this library already uses aws-sdk/client-s3

PR #504, opened 2023-10-22
@tus/s3-store: Fix invalid chunk size for uploads with deferred length

### LINKED ISSUE #503: @tus/s3-store invalid calculated chunk size if upload size is deferred
If size is deferred (`size === undefined`), calculated chunk size in `NaN`, which is then passed to `StreamSplitter`. I'm not sure of actual side effects beyond "undefined behavior", but my best guess is that stream is not split and entire stream is buffered to disk before being uploaded as single part to S3.

https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L272-L272

This was probably overlooked because of unsafe cast to `number` which disables static checks.
https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L265

Edit: I have verified that stream in not split which results in a single part upload. This limits the maximum upload size to 5GB which is a upper limit for part size. 

Easy solution would be to set the `chunkSize` to 5GB is this case. The downside is that we would always cache up to 5GB before part upload would event start but that is also true for any other very (very) large file where file size is known beforehand.

mitjap:
This PR fixes #503.

When upload size is not known current calculated chunk size is `NaN` which causes request to be cached to disk without upper limit. Later it would be uploaded as part. If request was larger than 5GB (max size of single part) the upload would fail. 

This fix assumes maximum possible file size (5TiB) when calculating chunk size. Calculated chunk size for this size is `~524MiB`.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -269,8 +269,6 @@ export class S3Store extends DataStore {
     let bytesUploaded = 0
     let currentChunkNumber = 0
 
-    console.log("this.calcOptimalPartSize(size)", this.calcOptimalPartSize(size));
-
     const splitterStream = new StreamSplitter({
       chunkSize: this.calcOptimalPartSize(size),
       directory: os.tmpdir(),
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -18,7 +18,7 @@ function calcOffsetFromParts(parts?: Array<AWS.Part>) {
 }
 
 type Options = {
-  // The preferred part size for parts send to S3. Can not be lower than 5MB or more than 500MB.
+  // The preferred part size for parts send to S3. Can not be lower than 5MB or more than 5GB.
   // The server calculates the optimal part size, which takes this size into account,
   // but may increase it to not exceed the S3 10K parts limit.
   partSize?: number
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -18,7 +18,7 @@ function calcOffsetFromParts(parts?: Array<AWS.Part>) {
 }
 
 type Options = {
-  // The preferred part size for parts send to S3. Can not be lower than 5MB or more than 5GB.
+  // The preferred part size for parts send to S3. Can not be lower than 5MiB or more than 5GiB.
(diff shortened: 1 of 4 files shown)


packages/s3-store/index.ts
```diff
@@ -70,7 +70,8 @@ export class S3Store extends DataStore {
   private client: S3
   private preferredPartSize: number
   public maxMultipartParts = 10_000 as const
-  public minPartSize = 5_242_880 as const // 5MB
+  public minPartSize = 5_242_880 as const // 5MiB
+  public maxUploadSize = 5_497_558_138_880 as const // 5TiB
```
Murderlon, line 74: We are confusing the max total size vs the max part size here I think

```suggestion
  public maxPartSize = 5_368_709_120  as const // 5GiB
```

Murderlon (commented):
Makes sense!

packages/s3-store/index.ts
```diff
@@ -422,7 +423,12 @@ export class S3Store extends DataStore {
     this.cache.delete(id)
   }
 
-  private calcOptimalPartSize(size: number): number {
+  private calcOptimalPartSize(size?: number): number {
+    // When upload size is not know we assume largest possible value (`maxUploadSize`)
+    if (size === undefined) {
+      size = this.maxUploadSize
```
Murderlon, line 429:
```suggestion
    // When upload size is not know we assume largest possible value
    if (size === undefined) {
      size = this.maxPartSize
```

mitjap: Variable `size` is total size of upload (`Upload-Length`), which can be deferred and thus undefined. In this case we can not assume much, but upper limit is still 5TB (max upload size). Using this `size` this function will calculate appropriate part size while also considering max part size and maximum number of parts.

This logic has some flaws but it works for general case and does not exceed max part size.

Murderlon: ah of course. I often glance too quickly (with dyslexia) and managed to think we were ending up with the part size of `maxUploadSize`. All good here 👍

PR #506, opened 2023-10-25
@tus/server: Split to multiple chunks on large incoming buffer

mitjap:
When handling incoming request, each buffer length is 64kb which `StreamSplitter` can handle without issues.

But in tests we might create buffers much larger than single part size, which might not be split as expected.

Works:
`const readStream = Readable.from(Buffer.alloc(expectedChunks * optimalChunkSize))`

Doesn't work:
`const readStream = Readable.from(`**`[`**`Buffer.alloc(expectedChunks * optimalChunkSize)`**`]`**`)`


Murderlon (approved):
👌

PR #507, opened 2023-10-26
@tus/s3-store: Fix incorrect part retreival procedure

mitjap:
This PR fixes some issues with `retrieveParts` function.

First issue is how it determines when to fetch more parts. Before it was using `data.NextPartNumberMarker && Number(data.NextPartNumberMarker) > 0` to check if there are more parts to be fetch. This code made an assumption that final "page" will return "0" as `NextPartNumberMarker` but this is not specified in documentation and differs between S3 providers. Scaleway's S3 for example always return last part's number. Specification is clear on this that we should use `IsTruncated` propery.

Next issue is with sorting the parts. When completing multipart upload parts must be ordered in ascending order. In existing code sorting was done only when there less then 1000 parts (no additional "pages").

Another issue is with filtering. I'm not sure why would we ever need to remove parts from this list.

### DIFF
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -383,33 +383,23 @@ export class S3Store extends DataStore {
       Bucket: this.bucket,
       Key: id,
       UploadId: this.cache.get(id)?.['upload-id'],
-    }
-    if (partNumberMarker) {
-      params.PartNumberMarker = partNumberMarker
+      PartNumberMarker: partNumberMarker,
     }
 
     const data = await this.client.listParts(params)
 
-    // INFO: NextPartNumberMarker should be undefined when there are no more parts to retrieve,
-    // instead it keeps giving `0` so to prevent an infinite loop we check the number.
-    if (data.NextPartNumberMarker && Number(data.NextPartNumberMarker) > 0) {
-      return this.retrieveParts(id, data.NextPartNumberMarker).then((parts) => {
-        if (parts && data.Parts) {
-          return [...data.Parts, ...parts]
-        }
-        return data.Parts
-      })
+    let parts = data.Parts ?? []
+
+    if (data.IsTruncated) {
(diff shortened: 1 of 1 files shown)


packages/s3-store/index.ts
```diff
@@ -490,7 +478,8 @@ export class S3Store extends DataStore {
     // Metadata request needs to happen first
     const metadata = await this.getMetadata(id)
     const parts = await this.retrieveParts(id)
-    const partNumber = parts?.length ?? 0
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const partNumber: number = parts.length > 0 ? parts[parts.length - 1].PartNumber! : 0
     const nextPartNumber = partNumber + 1
```
mitjap, line 483: `nextPartNumber` should be calculated from last part number

packages/s3-store/index.ts
```diff
@@ -378,38 +378,26 @@ export class S3Store extends DataStore {
...
+      const rest = await this.retrieveParts(id, data.NextPartNumberMarker)
+      parts = [...parts, ...rest]
+    }
 
-    if (parts && !partNumberMarker) {
-      return (
-        parts
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          .sort((a, b) => a.PartNumber! - b.PartNumber!)
-          .filter((value, index) => value.PartNumber === index + 1)
```
mitjap, line 411: Maybe this filter was kind of fail-safe if some part in the middle was not uploaded. This does not prevent later file corruption if some part was uploaded only partially.

Murderlon (approved):
Makes sense!

mitjap:
@Murderlon Would be possible to publish this fix?

Murderlon:
Done!
https://github.com/tus/tus-node-server/releases/tag/%40tus%2Fs3-store%401.1.1

Issue #459, opened 2023-07-19
Add `isCompleted` to `Upload` info class


Murderlon:
This could be useful for lifecycle rules in S3 to clean up incomplete parts and in general.

Acconut:
I am not sure if an entry inside the JSON content of the info object would help here (that's how I understand your proposal). AFAIK, S3 lifecycle rules are not able to access the content of files stored on S3 and can thus not filter out files that are missing the `isCompleted` property.

S3 tags might help here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html Each completed upload can get a `isCompleted` tag and then the lifecycle rules can filter out files that don't have this tag but are too old.

fenos:
Yes, this should be part of a tag in order for lifecycle rules to pick the change

fenos:
@mitjap incompleteParts are uploaded as real objects in S3 not as parts, hence this was required. However, i think that using the expiration date is cleaner as implemented here #513

Murderlon:
Closed in https://github.com/tus/tus-node-server/pull/513

PR #513, opened 2023-11-10
@tus/s3-store: add expiration extension

fenos:
Implements Expiration Extension for the S3 Store.

a new configuration is added on the S3Options to configure how long an upload URL will be valid for:

```ts
new S3Store({
 expirationPeriodInMilliseconds: ((1000 * 60) * 60) * 24, // 24h
})
```

### DIFF
--- a/packages/s3-store/README.md
+++ b/packages/s3-store/README.md
@@ -90,7 +90,7 @@ After a multipart upload is aborted, no additional parts can be uploaded using t
 
 ### Expiration
 
-Unlike other stores, the expiration extension on the S3 store does not need to call [`server.cleanExpiredUploads()`][cleanExpiredUploads].
+Unlike other stores, the expiration extension on the S3 store does not need to call [`server.cleanUpExpiredUploads()`][cleanExpiredUploads].
 The store creates a `Tus-Complete` tag for all objects, including `.part` and `.info` files, to indicate whether an upload is finished.
 This means you could setup a [lifecyle][] policy to automatically clean them up without a CRON job.
 
--- a/packages/s3-store/README.md
+++ b/packages/s3-store/README.md
@@ -79,7 +79,7 @@ The tus protocol supports optional [extensions][]. Below is a table of the suppo
 | ------------------------ | --------------- |
 | [Creation][]             | ✅              |
 | [Creation With Upload][] | ✅              |
-| [Expiration][]           | ❌              |
+| [Expiration][]           | ✅              |
 | [Checksum][]             | ❌              |
 | [Termination][]          | ✅              |
 | [Concatenation][]        | ❌              |
@@ -87,6 +87,32 @@ The tus protocol supports optional [extensions][]. Below is a table of the suppo
 ### Termination
 
 After a multipart upload is aborted, no additional parts can be uploaded using that upload ID. The storage consumed by any previously uploaded parts will be freed. However, if any part uploads are currently in progress, those part uploads might or might not succeed. As a result, it might be necessary to set an [S3 Lifecycle configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html) to abort incomplete multipart uploads.
+
+### Expiration
+
+Unlike other stores, the expiration extension on the S3 store does not need to call [`server.cleanExpiredUploads()`][cleanExpiredUploads].
+The store creates a `Tus-Complete` tag for all objects, including `.part` and `.info` files, to indicate whether an upload is finished.
+This means you could setup a [lifecyle][] policy to automatically clean them up without a CRON job.
+
+```json
+{
+  "Rules": [
+    {
+      "Filter": {
+        "Tag": {
+          "Key": "Tus-Complete",
+          "Value": "false"
+        }
+      },
+      "Expiration": {
+        "Days": 2
+      }
+    }
+  ]
+}
+```
+
+If you want more granularity, it is still possible to configure a CRON job to call [`server.cleanExpiredUploads()`][cleanExpiredUploads] yourself.
 
 ## Examples
 
@@ -137,3 +163,5 @@ See [`contributing.md`](https://github.com/tus/tus-node-server/blob/main/.github
 [checksum]: https://tus.io/protocols/resumable-upload.html#checksum
 [termination]: https://tus.io/protocols/resumable-upload.html#termination
 [concatenation]: https://tus.io/protocols/resumable-upload.html#concatenation
+[cleanExpiredUploads]: https://github.com/tus/tus-node-server/tree/main/packages/server#servercleanupexpireduploads
+[lifecyle]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -22,6 +22,7 @@ type Options = {
   // The server calculates the optimal part size, which takes this size into account,
   // but may increase it to not exceed the S3 10K parts limit.
   partSize?: number
+  expirationPeriodInMilliseconds?: number
   // Options to pass to the AWS S3 SDK.
   s3ClientConfig: S3ClientConfig & {bucket: string}
 }
@@ -69,6 +70,7 @@ export class S3Store extends DataStore {
   private cache: Map<string, MetadataValue> = new Map()
   private client: S3
   private preferredPartSize: number
+  private expirationPeriodInMilliseconds = 0
   public maxMultipartParts = 10_000 as const
   public minPartSize = 5_242_880 as const // 5MiB
   public maxUploadSize = 5_497_558_138_880 as const // 5TiB
@@ -82,9 +84,11 @@ export class S3Store extends DataStore {
       'creation-with-upload',
       'creation-defer-length',
       'termination',
+      'expiration',
     ]
     this.bucket = bucket
     this.preferredPartSize = partSize || 8 * 1024 * 1024
+    this.expirationPeriodInMilliseconds = options.expirationPeriodInMilliseconds ?? 0
     this.client = new S3(restS3ClientConfig)
   }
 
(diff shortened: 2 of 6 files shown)


fenos:
I've changed slightly the implementation, instead of relying on the `LastModified` returned by S3 which will change whenever we overwrite the metadata file (for example in the declareLength on deferred uploads) we add an additional metadata field `created-at` 

Additionally, if we configure an expiration for the upload, we can safely set the expiration for `incompleteParts` so that they will be automatically cleaned up once the URL "expires" which fixes #459

packages/s3-store/index.ts
```diff
@@ -30,6 +31,7 @@ type MetadataValue = {
   file: Upload
   'upload-id': string
   'tus-version': string
+  'created-at': string
```
Murderlon, line 34: Maybe I'm misunderstanding something but `file` is `Upload` which already has `creation_date`? Can't we reuse that?

Murderlon (commented):
Thank you!

I'm not sure this works currently as the store doesn't implement `deleteExpired`:

https://github.com/tus/tus-node-server/blob/612ac24cf7507ddeda57cab90f3bc95d9dfbcb6a/packages/server/src/server.ts#L232-L238

fenos:
@Murderlon @Acconut Unfortunately the `Expires` option during upload doesn't set the expiration to an object as i thought (it would have been too easy and nice) but it instead instruct the cache header when it should expire 😢 

For this reason, I've made quite a few changes as follows: 

- Re-introduced the concept of `Tus-Completed` tag in order to create a lifecycle policy to automatically clean up expired files. I'm using `putObject` instead of `putObjectTagging` just because the `.info` file is quite small and it has better compatibility across S3 servers (like R2)

- Implemented `deleteExpired` method on S3Store which relies on `listMultipartUploads` to list in-progress / unfinished uploads. For each expired `multiPartUpload` we delete `.info` - `.part` and abort the upload. One caveat is that MiniO doesn't honor the S3 server response of `listMultipartUploads` and returns an empty list see the issue https://github.com/minio/minio/issues/13246 - this approach wouldn't work with MiniO because of that not sure how to go about this peculiarity. Other S3 compatible stores such as R2 seem to return honour the S3 spec.

- Added e2e tests for the S3Store against a MiniO server to cover all expiration use cases. I do mock the `listMultipartUploads` for the reason reported above since I'd prefer to keep the proper compatibility on S3

Let me know how it feels now 👍

Acconut:
I am 👍 for using S3 lifecycle configurations to support expiration. That is the most efficient way to implement and it scales well for huge buckets, although at the cost of a reduced granularity (expiration can only be configured in days, not hours or minutes).

Could you please provide an example for the lifecycle configuration that is intended to be used with your implementation from this PR?

> * Implemented `deleteExpired` method on S3Store which relies on `listMultipartUploads` to list in-progress / unfinished uploads.

Why is it necessary to delete these manually? We can use the AbortIncompleteMultipartUpload rule (https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleRule.html#AmazonS3-Type-LifecycleRule-AbortIncompleteMultipartUpload) to remove the incomplete multipart uploads and then use "normal" expiration to clean up .info and .part files.

Acconut:
I agree with you here, but I don't understand why we need the `deleteExpired` implementation? Is it not possible to let S3 lifecycles clean up the multipart uploads for us? Why do we need to do this manually?

Murderlon:
@Acconut as mentioned in https://github.com/tus/tus-node-server/pull/513#pullrequestreview-1739511908 that's why API we currently have. Not having that method would result in a crash. It's also possible to have a dummy method which does nothing, but why not keep the API consistent across stores for CRON jobs?

packages/s3-store/index.ts
```diff
@@ -588,4 +612,91 @@ export class S3Store extends DataStore {
...
+          )
+        }) || []
+
+      const objectsToDelete = expiredUploads.reduce((all, expiredUpload) => {
+        all.push(
+          {
+            key: expiredUpload.Key + '.info',
+          },
+          {
+            key: expiredUpload.Key + '.part',
```
Murderlon, line 656: What do you think about adjusting/renaming `this.partKey()` to handle the extensions? Or alternatively, having to constants for the extensions and removing `partKey()`?

packages/s3-store/index.ts
```diff
@@ -588,4 +612,91 @@ export class S3Store extends DataStore {
...
+          }
+        )
+        return all
+      }, [] as {key: string}[])
+
+      const deletions: Promise<AWS.DeleteObjectsCommandOutput>[] = []
+
+      // Batch delete 1000 items at a time
```
Murderlon, line 664: Isn't a 1000 concurrent connection going to be rough over a HTTP/1.1 connection, which is AFAIK what S3 still uses?

Murderlon (commented):
Thanks a lot for this big effort! Testing locally with minio is pretty cool too.

I tried to add docs  but I can't push to this PR. Here is a patch you can use with `git apply 0001-Add-docs.path`

[0001-Add-docs.patch](https://github.com/tus/tus-node-server/files/13448642/0001-Add-docs.patch)

Acconut:
> as mentioned in [#513 (review)](https://github.com/tus/tus-node-server/pull/513#pullrequestreview-1739511908) that's the API we currently have. Not having that method would result in a crash. It's also possible to have a dummy method which does nothing, but why not keep the API consistent across stores for CRON jobs?

Yes, I am suggesting to just have a dummy method without doing any manual labor. Then the S3 lifecycles can take care of of this automatically without any cron job needed.

Maybe I am confused here. Is the cron implementation an alternative to using S3 lifecycles? Or do I need to run the cron implementation next to lifecycles in order to clean up all expired uploads (complete and incomplete)?

fenos:
> Maybe I am confused here. Is the cron implementation an alternative to using S3 lifecycles? Or do I need to run the cron implementation next to lifecycles in order to clean up all expired uploads (complete and incomplete)?

To clarify, I think that we can say we support both!
It's ultimately the user's choice if they want to rely on S3 lifecycle or call `cleanUpExpiredUploads()` in a cronJob it's either one or the other.

Even if i don't think it would cause any issue if the user has both lifecycle and cronjob but it doesn't really make sense :)

Acconut:
Alright, I thought users were expected to use both methods. Thanks for clarifying this.

However, the current cron implementation is not able to clean completed uploads, is it? I can imagine that this is also desired by some users.

Murderlon:
> Alright, I thought users were expected to use both methods. Thanks for clarifying this.

I clarified this in the docs git patch I shared above

> However, the current cron implementation is not able to clean completed uploads, is it? I can imagine that this is also desired by some users.

According the [spec](https://tus.io/protocols/resumable-upload#expiration):
> The Server MAY remove **unfinished** uploads once they expire

So might be out of scope?

packages/s3-store/index.ts
fenos: No, we send one request with at most 1000 objects as per S3 limitation.
In case there are 1M objects expired we would send 1000 requests with each 1000 objects. 

However as far as i know the AWS SDK ships with a default connection pool of 50, so in theory should never exceed [50 requests simultaneously](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/node-http-handler/src/node-http-handler.ts#L77) at any point, in order to increase or decrease this amount, you will need to pass a custom requestHandler:

```ts
const s3 = new S3Client({
  // Use a custom request handler so that we can adjust the HTTPS Agent and
  // socket behavior.
  requestHandler: new NodeHttpHandler({
    httpsAgent: new Agent({
      maxSockets: 500,

      // keepAlive is a default from AWS SDK. We want to preserve this for
      // performance reasons.
      keepAlive: true,
      keepAliveMsecs: 1000,
    }),
    socketTimeout: 5000,
  }),
});
```

Acconut:
> So might be out of scope?

The handling of completed uploads is not discussed much in the spec because that is mostly out of the scope of tus. That being said, tus servers often implement features that go beyond the content of the specification because those are needed to properly use and run tus severs, e.g. hooks for integrating with the main application. So deciding what is in scope or not is a blurry line :)

I think people will be interested in removing completed uploads as well since this will trim the bucket and keep it from growing indefinitely. Of course, you can decide that this is not something you want to handle in tus-node-server. However, if I were to implement expiration in tusd, I will handle this case as well.

packages/s3-store/index.ts
Murderlon: I read over the code a bit too quickly, this makes sense. Thanks.

fenos:
It is not trivial to add per Object expiration (for successful uploads), the reason is that you have no other way than to iterate the whole bucket keys and compare the LastModified date in S3 with a date you set as the expiration.

I believe we can add this functionality in another iteration if needed to not increase the scope of this current PR too much,
another PR another problem to solve 😄

Murderlon:
I'm okay with keeping it only for unfinished uploads for now. We can revisit if the need arises.

Regarding minio, I thought about it more and I think it's probably better to reuse the existing S3 credentials in CI and actually test against s3. This would make it consistent with the other tests, we avoid the addition of docker in the repository, and we can get rid of the mocking for `listMultipartUploads`.

What do you think?

fenos:
Yes agree, I will use S3 credentials

fenos:
@Murderlon I've switched the tests to use the S3 bucket with the credentials provided as envs,
However, the test that calls `deleteExpired()` returned me 120 file deleted instead of 2 😄 probably there were some unfinished uploads on the bucket.

I'm not sure if this was a one off thing or not, could you try to re-run the tests pls, i don't have the power to trigger a re-run.
If it still fails, we'll need to find another way to assert the correctness of the test

packages/s3-store/index.ts
fenos: Done

fenos:
@Murderlon right! I've modified the test slightly to make sure it tests the right thing even if there are unfinished uploads (probably caused by other tests)

The test does the following:

- Upload 6MB out of 11MB
- Call .deleteExpired()
- Make sure the .info file and the .part file are still there and not deleted
- wait until the files should be expired
- re-run .deleteExpired()
- asserts that these 2 files are not there 👍

packages/s3-store/README.md
```diff
@@ -88,6 +88,32 @@ The tus protocol supports optional [extensions][]. Below is a table of the suppo
 
 After a multipart upload is aborted, no additional parts can be uploaded using that upload ID. The storage consumed by any previously uploaded parts will be freed. However, if any part uploads are currently in progress, those part uploads might or might not succeed. As a result, it might be necessary to set an [S3 Lifecycle configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html) to abort incomplete multipart uploads.
 
+### Expiration
+
+Unlike other stores, the expiration extension on the S3 store does not need to call [`server.cleanExpiredUploads()`][cleanExpiredUploads].
```
Murderlon, line 93: I made a typo, link is correct

```suggestion
Unlike other stores, the expiration extension on the S3 store does not need to call [`server.cleanUpExpiredUploads()`][cleanExpiredUploads].
```

Murderlon (approved):
Great! Thanks!

fenos:
I think is ready to be merged! 🎉 👍

PR #514, opened 2023-11-10
@tus/server: add support for lockers

### LINKED ISSUE #408: Add support for lockers, including a distributed lock
The protocol [states](https://tus.io/faq.html#how-do-i-scale-tus) that: 

In order to horizontally scale a tus server there should be locks in place during the time the server interacts with the remote storage (s3 for example)

Does tus-node-server already solve this issue in some way? 
If not, do we need to implement an interface to add these custom locks?

Thanks!

fenos:
Adds distributed locking capabilities via the `Locker` interface.

```ts
new Server({
   ...options,
   locker: new MemoryLocker(),
})
```

Useful to implement custom lockers with redis / postgres

closes #408


mitjap (changes requested):
Thank you for this PR. I gave it a quick look but I see an issue with `MemoryLocker`. When unlocking a lock, all pending locks are resolved.

```
it('unlocking a lock should allow only one pending lock to resolve', async () => {
  const locker = new MemoryLocker()
  const lockId = 'upload-id-1'

  await locker.lock(lockId)
  const lock2 = locker.lock(lockId)
  const lock3 = locker.lock(lockId)
  const lock4 = locker.lock(lockId)
  const lock5 = locker.lock(lockId)

  await locker.unlock(lockId)
  await lock2
  await lock3
  await lock4
  await lock5

  throw new Error('should not be reached')
})
```

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,34 @@
...
+      this.locks[id] = true
+      return Promise.resolve()
+    }
+
+    return new Promise((resolve) => {
+      this.lockEvent.once(`unlock:${id}`, () => {
+        this.locks[id] = true
+        resolve()
+      })
+    })
```
mitjap, line 27: On unlock event you need to check if lock is available. Only first pending lock can perform a lock. Others need to resubscribe to event.

```suggestion
    return new Promise((resolve) => {
      const callback = () => {
        if (!this.isLocked(id)) {
          this.locks[id] = true
          resolve()
        } else {
          this.lockEvent.once(`unlock:${id}`, callback)
        }
      }
      this.lockEvent.once(`unlock:${id}`, callback)
    })
```

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,34 @@
+import {EventEmitter} from 'node:events'
+
+export interface Locker {
+  lock(id: string): Promise<void>
+  unlock(id: string): Promise<void>
+}
+
+export class MemoryLocker implements Locker {
+  private lockEvent = new EventEmitter()
+  private locks: Record<string, boolean> = {}
```
mitjap, line 10: Let's make this a `Map<string, boolean>`.

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  isLocked(id: string) {
+    return Boolean(this.locks[id])
```
mitjap, line 13:
```suggestion
    return this.locks.get(id) ?? false
```

packages/server/test/Locker.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    await locker.lock(lockId)
+    setTimeout(() => {
+      locker.unlock(lockId)
+    }, 300)
+    await locker.lock(lockId) // will wait until the other lock is released
+    await locker.unlock(lockId)
+    const endDate = new Date().valueOf() - date.valueOf()
+    assert.equal(endDate >= 300, true)
+  })
+})
```
mitjap, line 19: Here are some more tests that I recommend we add.

```suggestion
  it('locking a locked lock should not resolve', async () => {
    const locker = new MemoryLocker()
    const lockId = 'upload-id-1'

    await locker.lock(lockId)

    const p1 = locker.lock(lockId)
    const p2 = new Promise((resolve) => setTimeout(resolve, 500, 'timeout'))

    assert.strictEqual(await Promise.race([p1, p2]), 'timeout')
  })

  it('unlocking a lock should resolve only one pending lock', (done) => {
    const locker = new MemoryLocker()
    const lockId = 'upload-id-1'

    ;(async () => {
      await locker.lock(lockId)
      locker.lock(lockId).then(done)
      locker.lock(lockId).then(() => {
        done(new Error())
      })
      locker.lock(lockId).then(() => {
        done(new Error())
      })
      locker.lock(lockId).then(() => {
        done(new Error())
      })
      await locker.unlock(lockId)
    })()
  })

  // NOTE: does this test make sense?
  it('unlocking a lock should first resolve unlock promise and then pending lock promise', async () => {
    const locker = new MemoryLocker()
    const lockId = 'upload-id-1'

    await locker.lock(lockId)

    const resolveOrder = new Array<number>()
    const p1 = locker.lock(lockId).then(() => {
      resolveOrder.push(2)
    })
    const p2 = locker.unlock(lockId).then(() => {
      resolveOrder.push(1)
    })

    await Promise.all([p1, p2])
    assert.deepStrictEqual(resolveOrder, [1, 2])
  })
})
```

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,34 @@
...
+        resolve()
+      })
+    })
+  }
+
+  async unlock(id: string): Promise<void> {
+    delete this.locks[id]
+    this.lockEvent.emit(`unlock:${id}`)
+  }
+}
```
mitjap, line 34: My version of `MemoryLocker` with queue of pending locks, where only one is notified on unlock.

```suggestion
class Lock {
  public locked = false
  public pendingLocks = new Array<() => void>()
}

export class MemoryLocker implements Locker {
  private locks = new Map<string, Lock>()

  private getLock(id: string): Lock {
    let lock = this.locks.get(id)
    if (lock === undefined) {
      lock = new Lock()
      this.locks.set(id, lock)
    }
    return lock
  }

  async lock(id: string): Promise<void> {
    const lock = this.getLock(id)

    if (lock.locked) {
      await new Promise<void>((resolve) => {
        lock.pendingLocks.push(resolve)
      })
    }

    lock.locked = true
  }

  async unlock(id: string): Promise<void> {
    const lock = this.getLock(id)
    if (!lock.locked) {
      throw new Error('Releasing an unlocked lock!')
    }

    lock.locked = false

    // notify next pending lock
    const callback = lock.pendingLocks.shift()
    if (callback === undefined) {
      this.locks.delete(id)
    } else {
      callback()
    }
  }
}

```

fenos:
Thanks a lot for the great review!
I've changed the memory lock with the queue implementation you proposed! neat!

Let me know if now looks good!

One more note on this, is that - I think we might want to optionally allow to provide a function in to create the locker class.
In my case I run TUS on a multi-tenant way and I'd need to connect to the right backend to acquire the lock.

For example:

```ts
new Server({
   ...options,
   locker: (req: Request) => new MyCustomLocker(req.db),
})
```

then in the BaseHandler i have a getter method:

```ts
getLocker(req: http.IncomingMessage) {
   if (typeof this.locker === 'function') {
      return this.locker(req)
   } 
   return this.locker
}
```

and calling it this way:

```ts
const locker = this.getLocker(req)
try {
  locker?.lock(id)
} finally {
  locker?.unlock(id)
}
```

Any thoughts on this?

Acconut (commented):
This is a great PR, thanks for working on it! Locks are a good mechanism for taming the concurrent nature of requests for resumable uploads, but they also come with their own challenges. I have described this in more detail in tusd's documentation at https://github.com/tus/tusd/blob/main/docs/locks.md#avoiding-locked-uploads. For tusd, we have also implement a mechanism to unlock uploads if a new request attempts to resume the upload. Something similar should be considered here as well.

> I think we might want to optionally allow to provide a function in to create the locker class.
In my case I run TUS on a multi-tenant way and I'd need to connect to the right backend to acquire the lock.

I see where this could be useful, even though I don't have a use case and thus cannot judge if this approach comes with caveats. However, if we were to introduce multi-tenant support for lockers, this should be done in a uniform way that also works for the storages, for example.

Acconut:
> Do you guys have a way to do it with maybe with Redis?

Redis has a great pub/sub mechanism to send messages. The lock holder could listen for notifications about requests to release the lock. Other request handlers that are interested in acquiring the lock can instruct the holder to cleanup and release the lock by sending a message to a channel which is identified through the upload ID.

> I think it makes sense to also optionally support this, by simply creating a locker instance per request, since the locking mechanics are dealt with by the database itself

I don't disagree with this approach, but it's better to think about multi-tenancy for all features of tus-node-server versus building an approach that is just designed for lockers. That ensures that the API is consistent and usable for the long-term future :)

Murderlon:
> Redis has a great pub/sub mechanism to send messages. The lock holder could listen for notifications about requests to release the lock

Using plain pub/sub for locks might be dangerous as there is no guaranteed arrival. This is why [RedLock](https://redis.io/docs/manual/patterns/distributed-locks/#the-redlock-algorithm) exists for Redis, which is what I used in my experiments with distributed locks in tusd. 

> I don't disagree with this approach, but it's better to think about multi-tenancy for all features of tus-node-server versus building an approach that is just designed for lockers. That ensures that the API is consistent and usable for the long-term future :)

I think I agree that taking a step back and seeing how this related to other tus features would be best. But that's a lot more work and likely a breaking change. Since this only adds the ability to turn the locker optionally into a function, I would be okay with merging it as is to not block the PR and the use case.

Murderlon:
> For tusd, we have also implement a mechanism to unlock uploads if a new request attempts to resume the upload. Something similar should be considered here as well

I do think that we should explore this before merging though.

Acconut:
> Using plain pub/sub for locks might be dangerous as there is no guaranteed arrival.

I am not advocating the use of pub/sub for locking itself, but as a mechanism to communicate that another request handler is interested in acquiring the lock. RedLock might be the best way to implement the lock and then use pub/sub to signal that the lock holder should release it.

> Since this only adds the ability to turn the locker optionally into a function, I would be okay with merging it as is to not block the PR and the use case.

Alright, I have no strong opinion on this. Feel free to go that route :)

fenos:
I currently rely on Postgres Advisory locks to acquire the locks, which works great! However, the complicated part is to signal other servers when to release the lock in a distributed fashion!

Redis Pub/Sub might work in this case 👍

fenos:
Hello @Acconut @Murderlon @mitjap I've reworked the implementation in order to support the request cancellation as per spec.

In order for this to work reliably I needed to introduce an AbortController (something like the `Context` in go) and propagate its signal to all AWS functions as well as all the streams, to cancel their execution.

Once an operation is canceled, the abort controller will propagate the error `ERRORS.ABORTED` where the server implementation will then respond with `Connection: close` and terminate the connection early

Let me know if this sounds good to you 👍

Acconut:
> In order for this to work reliably I needed to introduce an AbortController (something like the `Context` in go) and propagate its signal to all AWS functions as well as all the streams, to cancel their execution.

FYI, the implementation in tusd works lightly different. Instead of sending an abort signal that is relayed to the S3 SDK, we effectively stop the incoming request body, so that no more data is read. For the stores, it effectively looks like the request body just ended naturally.

The benefit is that stores do not have to implement an abort mechanism (which is here the case AFAIK). In addition, passing the abort signal to the S3 SDK will likely cause the current operation to fail, meaning that transmitted data might get lost and has to be reuploaded. Stopping the request body appears like a normal EOF to the store, so it will just save all data normally.

I am not sure if a similar approach is feasible in tus-node-server but I just wanted to share the details on how tusd implements it.

fenos:
Hi @Acconut thanks for sharing this!

However, since you pass the `ctx` into all the Store functions I thought that if the request is canceled subsequently the context is canceled and also the store operations are then canceled.

Are you sure this is not the case in Tusd?

During upload, if I abort reading the request body (calling req.destroy()) it would indeed propagate the error to the stores, however, this is only true when an upload is ongoing.

For other operations, such as HEAD / DELETE / POST, we don't really read the request body so canceling the request would not make any real difference, thoughts on this?

Acconut:
> However, since you pass the `ctx` into all the Store functions I thought that if the request is canceled subsequently the context is canceled and also the store operations are then canceled.
> 
> Are you sure this is not the case in Tusd?

If a lock is requested to be released, the context will make sure to first close the request body. After a delay (a few seconds usually), the context will be cancelled so that all operations will halt. The idea is that closing the request body is enough to cause the storage to exit without error. But if an operation is hanging, the context cancellation will take care of it.

> For other operations, such as HEAD / DELETE / POST, we don't really read the request body so canceling the request would not make any real difference, thoughts on this?

These operations should usually finish quickly, so we can just wait until the lock is released.

fenos:
@Acconut Awesome Thanks!
I've got it working as you explained! 

To be honest is a bit hacky 😄 because node doesn't provide a nice way to cancel the stream in a graceful way, for this reason, I've created an `InteruptableStream` which listens for the abort signal. Once the signal is aborted I'm instructing the consumer that there is no more data (by pushing null as the last part).

During the response, I check if the context was canceled and destroy the request (which closes the connection)

I've included a test that simulates this behavior!

Acconut:
That's great to hear 🎉

fenos:
@Acconut @Murderlon doing latest tests on this, will update when everything is fully ready

fenos:
Hi @Acconut @Murderlon 

I've recorded the behavior of Locks with tus-node-server and tusd to compare, please see below.
I've noticed a subtle difference which I can't really figure out why tusd is behaving as follows:

As you can see in the video when the first HEAD (on the right tab) is issued, it will cancel the PATCH request as expected, however, i would have expected that it would be able to acquire the lock right away. 
Instead, it somehow doesn't and lets the other HEAD request be sent right after (somehow acquiring the lock) where the HEAD request that canceled the PATCH will then hang 16s.

It seems that even if the HEAD request cancels the other request, it doesn't try to acquire it right away why is that? how does this mechanism work? 😄 

In the tus-node-server you can see that all requests cancel each other just fine and are able to grab the lock right away.
It is true that it can happen that 2 PATCH requests gets sent at the same time, but as you can see in the video when that happens a `409 offset conflict` is returned, but the performance seems better as the locks are released faster

## TusD Locks
https://github.com/tus/tus-node-server/assets/4754064/d1ab3da8-34c7-4bec-99d2-cdd904034db2

## Tus-Node-Server
https://github.com/tus/tus-node-server/assets/4754064/adadb393-c220-42dd-8f87-69e3feafc5cd



Any feedback?

Acconut:
Which options did you pass to tusd and what were the logs from tusd?

fenos:
I've started `tusd` simply with `./tusd -upload-dir=./data`

some of the logs:

```
2023/11/29 08:46:14.137300 level=INFO event=UploadCreated method=POST path="" requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 id=7b2d9fc9570b08e8fac7aa71bac69f45 size=1048576000 url=http://127.0.0.1:8080/files/7b2d9fc9570b08e8fac7aa71bac69f45
2023/11/29 08:46:14.137320 level=INFO event=ResponseOutgoing method=POST path="" requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=201 body=""
2023/11/29 08:46:14.156905 level=INFO event=RequestIncoming method=OPTIONS path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:14.157770 level=INFO event=ResponseOutgoing method=OPTIONS path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" status=200 body=""
2023/11/29 08:46:14.161706 level=INFO event=RequestIncoming method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:14.170573 level=INFO event=ChunkWriteStart method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 maxSize=1048576000 offset=0
2023/11/29 08:46:21.330795 level=INFO event=RequestIncoming method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:24.172936 level=INFO event=UploadInterrupted method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45
2023/11/29 08:46:24.174436 level=ERROR event=BodyReadError method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 error="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource"
2023/11/29 08:46:24.174457 level=INFO event=ChunkWriteComplete method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 bytesWritten=6209536
2023/11/29 08:46:24.174494 level=INFO event=ResponseOutgoing method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=400 body="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource\n"
2023/11/29 08:46:24.707056 level=INFO event=RequestIncoming method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:24.710584 level=INFO event=ResponseOutgoing method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=200 body=""
2023/11/29 08:46:24.719365 level=INFO event=RequestIncoming method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:24.721104 level=INFO event=ChunkWriteStart method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 maxSize=1042366464 offset=6209536
2023/11/29 08:46:29.722656 level=INFO event=UploadInterrupted method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45
2023/11/29 08:46:29.723624 level=ERROR event=BodyReadError method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 error="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource"
2023/11/29 08:46:29.723647 level=INFO event=ChunkWriteComplete method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 bytesWritten=3096576
2023/11/29 08:46:29.723702 level=INFO event=ResponseOutgoing method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=400 body="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource\n"
2023/11/29 08:46:30.240538 level=INFO event=RequestIncoming method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:30.242804 level=INFO event=ResponseOutgoing method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=200 body=""
2023/11/29 08:46:30.255722 level=INFO event=RequestIncoming method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:30.258215 level=INFO event=ChunkWriteStart method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 maxSize=1039269888 offset=9306112
2023/11/29 08:46:35.259346 level=INFO event=UploadInterrupted method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45
2023/11/29 08:46:35.261381 level=ERROR event=BodyReadError method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 error="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource"
2023/11/29 08:46:35.261409 level=INFO event=ChunkWriteComplete method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 bytesWritten=3096576
2023/11/29 08:46:35.261448 level=INFO event=ResponseOutgoing method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=400 body="ERR_UPLOAD_INTERRUPTED: upload has been interrupted by another request for this upload resource\n"
2023/11/29 08:46:35.365759 level=INFO event=ResponseOutgoing method=HEAD path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 status=200 body=""
2023/11/29 08:46:35.381366 level=INFO event=RequestIncoming method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId=""
2023/11/29 08:46:35.384545 level=INFO event=ChunkWriteStart method=PATCH path=7b2d9fc9570b08e8fac7aa71bac69f45 requestId="" id=7b2d9fc9570b08e8fac7aa71bac69f45 maxSize=1036173312 offset=12402688
```

fenos:
@Murderlon Ready for review!

fenos:
@Acconut I think I know what's going on here!
My implementation of the MemoryLock uses `setImmediate` which essentially is like an async queue.

For this reason, the first request that tries to acquire the lock will be guaranteed to acquire it once the lock is released from another request.

In Tusd things work slightly differently because of go routines, and the memory locker in Tusd makes use of Mutexes to guarantee atomicity while accessing and modifying the locks, but the implementation doesn't use a FIFO-queue-based approach, the fastest request to obtain the lock wins, even if the previous request was already in a pending state.

the use of `setImmediate` in node, pushes the execution at the end of the I/O cycle in a queue, once the cycle completes all the callbacks are called in the same order as they were pushed.

Because of Node event loop single core execution, an alternative to mimic a similar behaviour of TusD is to add a waiting timer each time the lock was tried to be acquired but was still locked.

```ts
let acquired = false
while(!acquired) {
  lock.releaseRequest()
  acquired = await this.acquireLock()
  
  if (acquired) {
    return
  }
  await wait(100)
}
```

Thoughts? CC: @Murderlon

packages/server/src/server.ts
```diff
@@ -202,17 +218,38 @@ export class Server extends EventEmitter {
...
       headers['Content-Length'] = Buffer.byteLength(body, 'utf8')
     }
+
+    if (isAborted) {
+      // @ts-expect-error not explicitly typed but possible
+      headers['Connection'] = 'close'
+      res.writeHead(status, headers)
+      res.write(body)
+      const sent = res.end()
+      req.destroy()
```
Murderlon, line 248: not sure if it matters but `res.end()` has a callback, shouldn't we call `destroy` there to avoid a race condition?

test/package.json
```diff
@@ -10,14 +10,16 @@
     "@tus/file-store": "workspace:^",
     "@tus/gcs-store": "workspace:^",
     "@tus/s3-store": "workspace:^",
-    "@tus/server": "workspace:^"
+    "@tus/server": "workspace:^",
+    "throttle": "^1.0.3"
```
Murderlon, line 14: this needs to be a dev dependency

packages/server/src/server.ts
fenos: Yes, good point

fenos: i'm now calling `req.destroy()` on the `res.on('finish')` event which is essentially the same thing as the callback

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,93 @@
+import {ERRORS} from '../constants'
+
+export type RequestRelease = () => Promise<void> | void
+
+export interface Locker {
```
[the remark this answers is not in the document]
fenos: added comments, let me know if this is what you had in mind

packages/server/src/server.ts
```diff
@@ -230,4 +267,37 @@ export class Server extends EventEmitter {
 
     return this.datastore.deleteExpired()
   }
+
+  protected createContext(req: http.IncomingMessage) {
```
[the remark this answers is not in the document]
fenos: added some docs on this

Acconut:
@fenos I looked into the tusd behavior you mentioned. If you use tusd with the default disk-based storage, it will not use an in-memory locking mechanism but the disk-based locker (see https://github.com/tus/tusd/blob/main/pkg/filelocker/filelocker.go). In this implementation, the requests for releasing a lock works by creating a specific file on disk. The lock holder polls for this file to see if it should wind down the request handling. The lock interested entity will poll to see if the lock was released. To reduce the performance impact of this polling, the default intervals are between 5 to 10s IIRC.

However, this also means that there is a delay between the lock being released and the lock being acquired by the routine that request the lock to be unlocked. During this delay, another request (e.g. a HEAD request) could acquire the lock again. As you demonstrated this your video, this could lead to a situation where two clients are competing for the upload without making good, overall progress.

However, I don't think this should be a problem in general because there should not be two active clients uploading to the same upload resource anyways. In practice, the feature for requesting lock releases is only used when one request is dead and abandoned for its clients. In those cases, no retries should be started for it anyways.

That being said, I am also curious to see if we can improve tusd's handling for locks. The logic for requesting lock releases is rather new anyways, so there is likely potential for improvement :)

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -3,11 +3,19 @@ import EventEmitter from 'node:events'
...
 import type {DataStore} from '../models'
 import type http from 'node:http'
+import stream from 'node:stream'
+import {ERRORS} from '../constants'
 
 const reExtractFileID = /([^/]+)\/?$/
 const reForwardedHost = /host="?([^";]+)/
 const reForwardedProto = /proto=(https?)/
 
+export interface CancellationContext {
```
Acconut, line 13: Can you please add documentation for this interface? What do the properties `abort` and `cancel` mean?

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -101,4 +110,58 @@ export class BaseHandler extends EventEmitter {
 
     return {host: host as string, proto}
   }
+
+  protected getLocker(req: http.IncomingMessage) {
+    if (typeof this.options.locker === 'function') {
+      return this.options.locker(req)
```
Acconut, line 116: Should we allow `this.options.locker` to be an `async` function? People will likely ask for this to do asynchronous lookups, for example.

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -101,4 +110,58 @@ export class BaseHandler extends EventEmitter {
...
+    req: http.IncomingMessage,
+    id: string,
+    context: CancellationContext
+  ) {
+    const locker = this.getLocker(req)
+    await locker?.lock(id, () => {
+      context.cancel()
+    })
+
+    return () => locker?.unlock(id)
```
Acconut, line 131:
```suggestion
    if (locker == null) return
    await locker.lock(id, () => {
      context.cancel()
    })

    return () => locker.unlock(id)
```

What about this? Then we can get rid of these null checks.

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -101,4 +110,58 @@ export class BaseHandler extends EventEmitter {
...
+        }
+      })
+
+      req.on('error', (err) => {
+        if (!proxy.closed) {
+          proxy.destroy(err)
+        }
+      })
+
+      this.store.write(req.pipe(proxy), id, offset).then(resolve).catch(reject)
```
Acconut, line 164: Could `stream.pipeline` be used here to simplify the error handling?

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export type RequestRelease = () => Promise<void> | void
+
+// The Locker interface defines methods for implementing a locking mechanism.
+// This is crucial for ensuring exclusive access to uploads and their metadata.
+// Following TUS recommendations, it's important to cancel locks from previous requests
+// to avoid holding locks for too long and to manage half-open TCP connections effectively.
+// The lock method includes a cancel callback to facilitate the cancellation of a request that previously acquired the lock.
+export interface Locker {
+  lock(id: string, cancelReq: RequestRelease): Promise<void>
```
Acconut, line 11: In tusd, we don't pass the upload ID directly to `lock`/`unlock`, but instead let the locker create a lock instance, which is then used for locking and releasing: https://github.com/tus/tusd/blob/c856f723619a0a388c619d68357b9510c8632640/pkg/handler/datastore.go#L163-L190

The idea is that the lock instance can then hold references to the lock object in external services. This might be more challenging if we only operate on the upload ID.

Acconut: It seems like such a lock instance is already used by the memory locker internally.

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,106 @@
...
+
+class Lock {
+  public requestRelease?: RequestRelease
+}
+
+// MemoryLocker is an implementation of the Locker interface, maintaining locks in memory.
+// It ensures exclusive access to upload resources by managing locks for each upload.
+// The lock() method ensures that any previously held lock is released before acquiring a new one.
+// Lock acquisition attempts will timeout based on the specified 'acquireLockTimeout' duration.
+export class MemoryLocker implements Locker {
```
Acconut, line 27: Can we move this class into its own file?

packages/server/src/models/Locker.ts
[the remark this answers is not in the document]
Acconut: The documentation should also include more details on how the callback for lock cancellation actually works. If I understand it correctly, `cancelReq` will be invoked whenever another request is interested in acquiring the lock. In this case, the lock holder should stop processing the request and then call `unlock`. Please add something like this if it's correct.

packages/server/src/models/Locker.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    if (!lock) {
+      const lock = new Lock()
+      this.locks.set(id, lock)
+      return lock
+    }
+
+    await lock.requestRelease?.()
+
+    const potentialNewLock = this.locks.get(id)
```
Acconut, line 65: Is this code around `potentialNewLock` actually necessary? Isn't the acquiring of the lock handled by the `this.acquireLock` call a few lines below?

packages/server/src/server.ts
```diff
@@ -202,17 +211,38 @@ export class Server extends EventEmitter {
...
+    const isAborted = context.signal.aborted
+
     if (status !== 204) {
       // @ts-expect-error not explicitly typed but possible
       headers['Content-Length'] = Buffer.byteLength(body, 'utf8')
     }
+
+    if (isAborted) {
+      // @ts-expect-error not explicitly typed but possible
+      headers['Connection'] = 'close'
```
Acconut, line 237: Please add a comment explaining what setting this header does. Does it instruct the http module to close the connection?

Acconut (commented):
Amazing work! 😍

packages/server/src/handlers/BaseHandler.ts
fenos: unfortunately not, since if we use stream.pipeline it would automatically call `req.destroy()` and we don't want that, we first want to return a response, then call req.destroy()

packages/server/src/handlers/BaseHandler.ts
fenos: Do you have a use case in mind in which it could be beneficial to have it async?
I can't think of one atm

packages/server/src/handlers/BaseHandler.ts
fenos: we still need a no-op function returned as the `unlock` otherwise it would break the logic on the handler when calling `unlock`.

I would rather sacrifice this method with null checks that cluttering the handlers with:

```
if (unlock) {
 try {
 
 ... logic
 } finally {
  await unlock()
 }
} else {
 ... logic
}
```

packages/server/src/models/Locker.ts
fenos: it's more of an implementation detail the behaviour is the same and i find the interface easier to implement by only having `lock/unlock`

packages/server/src/models/Locker.ts
fenos: Sure thing

packages/server/src/models/Locker.ts
fenos: Yes good point!

packages/server/src/server.ts
fenos: This header instruct the client that we are willing to close the connection and to not send anymore data

packages/server/src/handlers/BaseHandler.ts
fenos: Sure thing!

packages/server/src/handlers/BaseHandler.ts
fenos: that's why I'm using a `PassThrough` proxy and add to that stream the abortSignal. This is thought to prevent the pipeline calling `req.destroy()` which will absurdly close the connection

packages/server/src/handlers/BaseHandler.ts
fenos: I've done what you recommended as this would save a no-op call to the function when the locker is not provided.
On the handler side I've done the null check

`unlock?.()`

packages/server/src/handlers/BaseHandler.ts
fenos: Done

packages/server/src/models/Locker.ts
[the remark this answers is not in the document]
fenos: Done

packages/server/src/handlers/BaseHandler.ts
Acconut: Alright, thank you.

packages/server/src/models/Locker.ts
Acconut: Yes, it is easier to implement it by passing around upload IDs as strings. But that interface is also more restricting because there is no state is preserved between locking and unlocking. Any state that should be preserved must be stored in a central map by the locker itself, which is not ideal. Having a lock instance with its methods for locking and unlock resolves this problem because the state can be kept in the instance itself.

I think in the future we would run into issues with the current interface, but right now we have the possibility to define a more powerful and future-proof interface.

fenos: In any case, the locker needs to keep a map of ID -> Lock as is also implemented here:
https://github.com/tus/tusd/blob/main/pkg/memorylocker/memorylocker.go#L29

By having the Locker simply return an instance of `Lock` which then the Lock interacts with the Locker I find it more complicated and don't yet see the usefulness that can bring in the future

packages/server/src/handlers/BaseHandler.ts
fenos: Done

Acconut:
Thank you very much for making the changes ❤️ I will leave it up to @Murderlon which interface approach is better suited in this case.

> use a global `EventEmitter` for listening to external release events.

I am not sure I understand this idea properly. With this approach, would these release events still be propagated through whatever external locking server the application is using? For example, if I have a cluster of servers running tus-node-server and I want to use Redis for locks, the release request events should also be routed through Redis, so that every tus-node-server instance has a change of receiving the event. This would not work if we use an EventEmitter, whose scope is limited to one tus-node-server instance.

fenos:
@Acconut Awesome thanks!

Just to clarify, in the example above, I'm using a local `EventEmitter` just to forward down to the interested lock, the event received from Redis pubSub.

```ts
const eventEmitter = new EventEmitter()
RedisPubSub.on('lock_release, (uploadId) => {
   eventEmitter.emit(`release:${id}`
})
```

Since you would want to have 1 single redis subscription in each `tus-node-server` node in the cluster, but we might have many locks with different ids created.

By forwarding the event with the EventEmitter only the `Lock` that acquired the lock will be notified:

```ts
class Lock {

...

lock(release) {
   await this.acquireLock()
   this.events.on(`release:${this.id}`, () => release())
}
```

it is just a nice pattern to have each lock only listening for the release events they are interested.

If we have a cluster of 6 nodes, each node will receive the redis event as well as emitting the local EventEmitter event, however, only the node that has acquired the lock will act upon the local event, all the other nodes didn't have a handler registered so the EventEmitter event is simply discarded.

Just to clarify, I have removed the `createContext` approach which i previously added in a previous commit, i think we can introduce it later if needed.

The main reason i introduced it is because i didn't want to create a new instance of `MyCustomLocker` for every request since i wanted to establish the subsctiption to `redis` in there only once.

But it can be worked around with a `static` method or initialising the subscription first and then passing it down to the Locker.

Murderlon:
I'm happy you reverted the `createContext` option as it keeps the PR leaner and I wasn't convinced on the implementation yet. Now it seems to be exactly where we want it to be API wise. Going through it now for final checks.

Murderlon (commented):
👍 on the code comments, I think it will really help long-term maintenance.

If I understand correctly, we don't set a default locker and thus we support `acquireLock` to also return `undefined`. Wouldn't it be better to set a default locker in the server constructor? Then we have added safety by default, we don't have conditional types everywhere and thus no `lock?.unlock()`, AFAIK we can even remove the `p-queue` dependency from `FileConfigStore` as locks should prevent this now:

https://github.com/tus/tus-node-server/blob/f1a4ac374b2590b3220feb36161a55e4b2e06f77/packages/file-store/configstores/FileConfigstore.ts#L8-L12

fenos:
@Murderlon Sure, i can add the `MemoryLocker` as the default locker

fenos:
@Murderlon Done, i've added the `MemoryLocker` as the default locker on the `BaseHandler` since there it seemed the more logical place to have it and removed `p-queue` from the file-store

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -18,15 +23,20 @@ export class BaseHandler extends EventEmitter {
       throw new Error('Store must be defined')
     }
 
+    if (!options.locker) {
+      options.locker = new MemoryLocker()
+    }
```
Murderlon, line 28: I don't see a reason why we wouldn't do this in the server constructor as that's the place where we check options. We can do something like this to type it correctly:

```ts
  options: WithRequired<ServerOptions, 'locker'>

  constructor(options: ServerOptions & {datastore: DataStore})
```

Murderlon (approved):
One last comment and then we can merge this 🎉

packages/server/src/handlers/BaseHandler.ts
fenos: there are a couple of reasons why i decided to not doing it on the Server constructor:

- 1) it means we need to add the `MemoryLocker` on all the tests around the code base that tests the BaseHandler or handlers

- 2) The TS definition seems way more ugly since we'll need the following:

```ts
//server.ts

export class Server extends EventEmitter {
  datastore: DataStore
  handlers: Handlers
  options: WithRequired<ServerOptions, 'locker'>

  constructor(options: ServerOptions & {datastore: DataStore}) {
    super()

    if (!options) {
      throw new Error("'options' must be defined")
    }

    if (!options.path) {
      throw new Error("'path' is not defined; must have a path")
    }

    if (!options.datastore) {
      throw new Error("'datastore' is not defined; must have a datastore")
    }

    if (!options.locker) {
      options.locker = new MemoryLocker()
    }

    const {datastore, ...rest} = options
 
    this.options = rest as WithRequired<ServerOptions, 'locker'>
  }
 ...
}
```

```ts
// baseHandler.ts
export class BaseHandler extends EventEmitter {
  options: WithRequired<ServerOptions, 'locker'>
  store: DataStore

  constructor(store: DataStore, options: WithRequired<ServerOptions, 'locker'>) {
    super()
    if (!store) {
      throw new Error('Store must be defined')
    }

    this.store = store
    this.options = options
  }
}
```

to be more ergonomic the handlers should have the default value, otherwise you cannot instantiate them without passing the locker in.

if this is what we want, then i can modify it as above

Murderlon: Maybe I misunderstand but since you already made `locker` a required argument on the handlers, TS should already complain if you don't add them in tests? So moving that around wouldn't change anything there.

Indeed I don't think you can get rid of the `rest as WithRequired<ServerOptions, 'locker'>`, but what's more important I think is that that happens _once_, rather than every time you access `locker` somewhere in the program.

What do you think?

fenos: > Maybe I misunderstand but since you already made locker a required argument on the handlers, TS should already complain if you don't add them in tests? So moving that around wouldn't change anything there.

No, because the constructor argument of the BaseHandler still uses `ServerOptions` which doesn't have the `locker` property required, the property `options` in the BaseHandler options needs to be `required`, this is why we can set the default value in the constructor if not provided.

But anyhow, I'll make the change and keep the default on the server constructor

fenos:
@Murderlon I've now added the default option on the Server as per the latest commit 👍

Murderlon:
Thanks for sticking with it and doing quick iterations on the many rounds of feedback 👌

netdown:
@Murderlon Is it possible that the locker mechanism is broken? I'm working on a GCS locker (will submit a PR), and noticed that the handler immediately calls unlock after locking. I switched to MemoryLocker, put some logging to `lock` and `unlock` functions and started a long running upload (the client is Uppy). It seems that when starting the upload, lock was called and immediately unlocked, and the same happened when the upload finished. Or am I missing something?

Murderlon:
We have tests for it and it runs in production at Supabase so if something is not working we'd need a reproducible example or a failing test.

PR #516, opened 2023-11-10
@tus/server: add upload id to `onIncomingRequest`

fenos:
This PR makes the `onIncomingRequest` hook more useful by adding the `uploadId` as the 3rd parameter.
99% of cases you'd like to use this hook to perform authorization logic. However is always very common to allow check for permission specifically for certain upload-id.

For example:

- Can this user add chunks to this upload?
- Can this user view this asset?

By adding the upload-id we can implement custom logic on a per asset basis

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -153,12 +153,6 @@ export class Server extends EventEmitter {
       return this.write(res, status_code, body)
     }
 
-    try {
-      await this.options.onIncomingRequest?.(req, res)
-    } catch (err) {
-      return onError(err)
-    }
-
     if (req.method === 'GET') {
       const handler = this.handlers.GET
       return handler.send(req, res).catch(onError)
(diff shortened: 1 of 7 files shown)


packages/server/src/server.ts
```diff
@@ -153,12 +153,6 @@ export class Server extends EventEmitter {
       return this.write(res, status_code, body)
     }
 
-    try {
-      await this.options.onIncomingRequest?.(req, res)
-    } catch (err) {
-      return onError(err)
-    }
-
```
mitjap, line 161: What if we only add `id` to this callback? This way we don't need to edit every handler.
```
    try {
      const id = this.handlers.PATCH.getFileIdFromRequest(req)
      await this.options.onIncomingRequest?.(req, res, id !== false ? id : undefined)
    } catch (err) {
      return onError(err)
    }
```

fenos: I can do that - the only thing is that it will be called twice this way

fenos: Also, this approach wouldn't work for the POST endpoint, which creates the URL

Murderlon (approved):
I think I'm okay with duplication here instead of trying to make the (wrong) abstraction. Do others feel strongly?

fenos:
What's the status on this? @mitjap @Murderlon 
I agree that instead of doing the wrong abstraction we can call them in all handlers.

If we ever are going to restructure the code at some point we can move them in a centralised place

PR #517, opened 2023-11-13
@tus/server: support `Tus-Max-Size`

fenos:
Based on the Tusd implementation, this PR implements a configurable maxSize option which will limit and validate incoming file size.

- We validate that the `upload-length` is not higher than the `maxSize` option provided.
- We limit the request body accordingly as in the Tusd implementation https://github.com/tus/tusd/blob/1a43e26f16f43bed5dd2219c27e6eb14c125fb03/pkg/handler/unrouted_handler.go#L810-L834

The only addition I've made is that in tusd it doesn't seem are limiting the request body to the maxSize specified.
They simply validate the upload-length header against the maxSize. 

However, in this PR if the body reaches the maxSize bytes read, it will throw a 413 error and stop uploading.
Not sure about the consequences as in what happens to the part of the file that has been already uploaded? 

In S3 we can set a policy to delete uncompleted multi-part but what about for fileStore?


Murderlon (commented):
Thanks a lot!

There is still some docs missing on this new option.

Acconut (commented):
Thanks to @fenos, we discovered a flaw in tusd's logic around handling size limits, so I would be a bit hesitant to copying its logic for now until we fixed that bug: https://github.com/tus/tusd/issues/1032

Has this code been tested extensively with requests that use `Transfer-Encoding: chunked`? In these cases, no `Content-Length` is set and the server must keep track of the read bytes and stop if necessary on its own.

fenos:
@Acconut I've implemented the bug fix on this PR, however will test more in depth the `Transfer-Encoding: chunked` in theory it should work with the current implementation

packages/server/src/models/StreamLimiter.ts
```diff
@@ -0,0 +1,34 @@
...
+  status_code: number
+  body: string
+
+  constructor() {
+    super(ERRORS.ERR_MAX_SIZE_EXCEEDED.body)
+    this.status_code = ERRORS.ERR_MAX_SIZE_EXCEEDED.status_code
+    this.body = ERRORS.ERR_MAX_SIZE_EXCEEDED.body
+    Object.setPrototypeOf(this, MaxFileExceededError.prototype)
+  }
+}
```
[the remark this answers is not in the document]
fenos: The reason I've created this error is that in the StreamLimiter the callback to signal an error must be an instance of `Error` as per typescript definition

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -78,4 +80,53 @@ export class BaseHandler extends EventEmitter {
...
+    }
+
+    if (length > 0) {
+      maxSize = length
+    }
+
+    // limit the request body to the maxSize if provided
+    if (configuredMaxSize > 0 && maxSize > configuredMaxSize) {
+      maxSize = configuredMaxSize
+    }
```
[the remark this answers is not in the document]
fenos: these are not mutually exclusive, the last statement makes sure that the maxSize never is never greater than the `configuredMaxSize`

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -78,4 +80,53 @@ export class BaseHandler extends EventEmitter {
...
+    if (typeof this.options.maxSize === 'function') {
+      return this.options.maxSize(req, id)
+    }
+    return this.options.maxSize ?? 0
+  }
+
+  async getBodyMaxSize(
+    req: http.IncomingMessage,
+    info: Upload,
+    configuredMaxSize?: number
```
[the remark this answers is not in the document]
fenos: Yes, I thought about this but we will need the `configuredMaxSize` outside of this function too (in the handler), For this reason instead of calling it twice (since it can be an async operation)

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -78,4 +80,53 @@ export class BaseHandler extends EventEmitter {
 
     return decodeURIComponent(match[1])
   }
+
+  getConfiguredMaxSize(req: http.IncomingMessage, id: string) {
+    if (typeof this.options.maxSize === 'function') {
+      return this.options.maxSize(req, id)
+    }
+    return this.options.maxSize ?? 0
```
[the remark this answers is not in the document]
fenos: maybe 50GB is a good enough default max size?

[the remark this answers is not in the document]
Murderlon: I think it's somewhat common for people to upload bigger than 50GB, but we can set a low default and prompt people to increase when needed 👍

packages/server/src/models/StreamLimiter.ts
```diff
@@ -0,0 +1,34 @@
+import {Transform, TransformCallback} from 'stream'
+import {ERRORS} from '../constants'
+
+export class MaxFileExceededError extends Error {
```
Murderlon, line 4:
```suggestion
// TODO: create HttpError and use it everywhere instead of throwing objects
export class MaxFileExceededError extends Error {
```

packages/server/test/PostHandler.test.ts
```diff
@@ -24,7 +25,10 @@ describe('PostHandler', () => {
   fake_store.hasExtension.withArgs('creation-defer-length').returns(true)
 
   beforeEach(() => {
-    req = {url: '/files', method: 'POST'} as http.IncomingMessage
+    req = new MockIncomingMessage({
+      url: '/files',
+      method: 'POST',
+    }) as unknown as http.IncomingMessage
```
Murderlon, line 31: I think we can replace this with `node-mocks-http`, which we already use:

https://github.com/tus/tus-node-server/blob/612ac24cf7507ddeda57cab90f3bc95d9dfbcb6a/packages/server/test/DeleteHandler.test.ts#L24

packages/server/src/handlers/BaseHandler.ts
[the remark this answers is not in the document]
fenos: @Murderlon I just realized that if we set a default it might be a breaking change for folks using tus-node-server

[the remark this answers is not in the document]
Murderlon: Good point. Question is at what point is it a breaking change and when is it a safety precaution? I think somewhere in the 200-500GB range we can justify it as the latter?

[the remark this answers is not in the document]
fenos: I think we can leave it as is, without a default - we only check `maxFileSIze > 0` in a couple of places, it's not that bad

fenos:
@Acconut @Murderlon I've made the following changes:

- I've rebased this PR 
- added a test when using `Transfer-Encoding: chuncked`
- Validate early when `Content-Lenght + offer > maxFileSize`
- Added documentation
- clean up code

Ready for a final review 🎉

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -158,7 +162,81 @@ export class BaseHandler extends EventEmitter {
         }
       })
 
-      this.store.write(req.pipe(proxy), id, offset).then(resolve).catch(reject)
+      stream
+        .pipeline(req.pipe(proxy), new StreamLimiter(maxFileSize), async (stream) => {
```
Acconut, line 166: Can we add a comment why we use `req.pipe(proxy)` here instead of `pipeline(req, proxy, new SteamLimiter, ...)`? I know you answered this question in another PR, but it's good to document this in the code as well for future encounters.

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -158,7 +162,81 @@ export class BaseHandler extends EventEmitter {
...
+    configuredMaxSize?: number
+  ) {
+    // Use the server-configured maximum size if it's not explicitly provided.
+    configuredMaxSize ??= await this.getConfiguredMaxSize(req, file.id)
+
+    // Parse the Content-Length header from the request (default to 0 if not set).
+    const length = parseInt(req.headers['content-length'] || '0', 10)
+    const offset = file.offset
+
+    const hasContentLengthSet = length > 0
```
Acconut, line 198: It is also possible to have `Content-Length: 0` set in a request. Maybe this check should rather be

```suggestion
    const hasContentLengthSet = req.headers['content-length'] != null
```

or someting similar.
```

File: packages/server/src/handlers/BaseHandler.ts [hunk 9]
```diff
@@ -158,7 +162,81 @@ export class BaseHandler extends EventEmitter {
...
+    }
+    return this.options.maxSize ?? 0
+  }
+
+  /**
+   * Calculates the maximum allowed size for the body of an upload request.
+   * This function considers both the server's configured maximum size and
+   * the specifics of the upload, such as whether the size is deferred or fixed.
+   */
+  async calculateMaxBodySize(
```
Acconut, line 186: This logic in this function appears a bit confusing to me. I wasn't able to spot a mistake, but the number of conditionals makes it hard to reason about.

I don't know exactly how it could be improved but maybe it's possible to refactor this into one overarching condition checking whether the size is deferred or not.

```
if file.sizeIsDeferred {

} else {

}
```

And then have the additional checks in its branches.

packages/server/README.md
```diff
@@ -60,6 +60,10 @@ Creates a new tus server with options.
 
 The route to accept requests (`string`).
 
+#### `options.maxSize`
+
+Max file size (in bytes) allowed when uploading (`number`)
```
Acconut, line 65: This needs documentation that `maxSize` can also be a function.

packages/server/src/handlers/OptionsHandler.ts
```diff
@@ -6,7 +6,13 @@ import type http from 'node:http'
 // A successful response indicated by the 204 No Content status MUST contain
 // the Tus-Version header. It MAY include the Tus-Extension and Tus-Max-Size headers.
 export class OptionsHandler extends BaseHandler {
-  async send(_: http.IncomingMessage, res: http.ServerResponse) {
+  async send(req: http.IncomingMessage, res: http.ServerResponse) {
+    const maxSize = await this.getConfiguredMaxSize(req, '')
```
Acconut, line 10: Please add documentation for the end users, that `options.maxSize` might be called with an empty ID. Or maybe it should not be an empty string but rather explicitly `null`.

Acconut (commented):
Nice work, @fenos!

packages/server/src/handlers/BaseHandler.ts
fenos: Done!

packages/server/src/handlers/BaseHandler.ts
fenos: Thanks! good catch!

packages/server/src/handlers/BaseHandler.ts
fenos: Cool! I agree!
I initially ported it as is from tusd, but I agree with your comment.

I have now cleaned it up - let me know if it looks good now

packages/server/README.md
fenos: Done

packages/server/src/handlers/OptionsHandler.ts
fenos: Done

fenos:
@Murderlon S3 tests are now failing with:

```
InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.
```

did you revoke the access key id?

packages/server/src/constants.ts
```diff
@@ -57,6 +57,14 @@ export const ERRORS = {
...
     body: 'The file for this url no longer exists\n',
   },
+  ERR_SIZE_EXCEEDED: {
+    status_code: 413,
+    body: "upload's size exceeded\n",
+  },
+  ERR_MAX_SIZE_EXCEEDED: {
+    status_code: 413,
+    body: 'Maximum size exceeded\n',
+  },
```
Murderlon, line 67: Can this be the same error? They seem very similar.

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -158,7 +161,71 @@ export class BaseHandler extends EventEmitter {
...
+    // Check if the upload fits into the file's size when the size is not deferred.
+    if (offset + length > (file.size || 0)) {
+      throw ERRORS.ERR_SIZE_EXCEEDED
+    }
+
+    if (hasContentLengthSet) {
+      return length
+    }
+
+    return (file.size || 0) - offset
```
Murderlon, line 229: I guess this is just TS shenanigans and we can never have a negative return value?

Murderlon (approved):
Looking great ✨ Thanks for the elaborate tests for this feature.

Only one small Q.

packages/server/src/handlers/BaseHandler.ts
fenos: yhea, `file.size` is an optional property so it can be undefined

packages/server/src/constants.ts
fenos: it can be yes, for some reason tusd seems to differentiate them

packages/server/src/handlers/BaseHandler.ts
fenos: but we are validating that the offset doesn't exceed the file.size just above 
https://github.com/tus/tus-node-server/pull/517/files#diff-e757cfc1d00bc5ab142584ea1fe7754ea3028e70a18262a3e926d998dbf78ca1R221

So it's fine to default to 0 to make TS happy

packages/server/src/constants.ts
fenos: MaxSizeExceeded: https://github.com/tus/tusd/blob/1a43e26f16f43bed5dd2219c27e6eb14c125fb03/pkg/handler/unrouted_handler.go#L31

SizeExceeded: https://github.com/tus/tusd/blob/1a43e26f16f43bed5dd2219c27e6eb14c125fb03/pkg/handler/unrouted_handler.go#L39

Technically i think they are 2 different errors:

- MaxSizeExceeded is returned when the body size is greater than the defined server max size
- SizeExceeded is returned when trying to upload more bytes than the declared size of the file 

I would personally keep them separate too

Issue #363, opened 2022-12-14
Introduce log levels


Murderlon:
Currently we use the `debug` package to log quite a few things. It would be a big improvement if we have a proper logger with log levels which can be configured as an option.

PR #518, opened 2023-11-20
feat: Custom error handler

fenos:
This PR introduces an optional onErrorResponse hook which allows to do useful things when an error occurs.

Example usage:

```ts
new Server({
 ...
 onErrorResponse: async (req, res, error, write) => {
   return write(error)
 }
})

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -147,9 +147,20 @@ export class Server extends EventEmitter {
       req.method = (req.headers['x-http-method-override'] as string).toUpperCase()
     }
 
-    const onError = (error: {status_code?: number; body?: string; message: string}) => {
+    const onError = async (error: {
+      status_code?: number
+      body?: string
+      message: string
+    }) => {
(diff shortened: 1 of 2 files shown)


Murderlon (commented):
I'm not sure on this change. We already allow custom errors in controlled circumstances in the `onUploadCreate` and `onUploadFinish` hooks, rather than _all_ errors. (Although I wish we have a `HttpError` model rather than an object). 

What's also important to note that as of now if post-processing fails in `onUploadFinish` and you send back a custom error, the client still considers it a success because the subsequent HEAD request will be successful.

What are some "useful" things you can do with errors? Logs could probably be solved with #363

Acconut:
I don't fully understand the use case for such a feature yet. Can you expand on what headers you would want to add in the case that an error occurs? Also, how should this work if multiple errors occur while handling a single request?

fenos:
I'll close this for now, I'm not entirely sure if I still need this hook, given that the other PRs would handle things properly now

Issue #24, opened 2016-05-19
GridFS support


mzygmunt:
Hello,
I'm currently working on MongoDB and I need to upload some large files into DB. Is there a plan to support GridFS?

bhstahl:
Sure, why not!

mhf-ir:
:+1:

PR #526, opened 2023-12-04
Add  GridFS support

spencerjibz:
### Background
Hi there, I have used gridFS for a while now, so recently thought might as well build a tus-store for it, I know there was a [related issue](https://github.com/tus/tus-node-server/issues/24) to this back in 2016.  

### Implementation Details
- [x]  Extended MongoDB's gridfs write stream to support updating existing files. (can contribute this back to the [mongo](https://github.com/mongodb/node-mongodb-native) repo later).
- [x]  Added a config Store for usage, internally used by the store package, but can also be used separately with other stores(FileStore).
- [x]  Modified the demos to include the new store.
- [x]  Added tests for the new store and also a MongoDB-in-memory server for tests and CI
- [x] Added e2e test for the gridfs-store 

Let me know if there are any changes you'd like me to make.


spencerjibz:
Just realised, l need to add e2e tests for the new store and also update readme.

Murderlon:
Sorry for the enormous delay. I really appreciate you taking the time to contribute. After thinking about it, I don't think taking on this maintenance burden and complexity is worth it for such a niche store. However I'll happily link this in the readme as a community plugin. Would you be interested in that?

PR #549, opened 2024-01-08
@tus/server: remove useless req.baseUrl & clarify docs

Murderlon:
Resolves https://github.com/tus/tus-node-server/issues/457#issuecomment-1879889301

- Remove `req.baseUrl`, which was `ts-ignore`'d, from `generateUrl`. This is always `undefined` and only exists in Express.js. From a type perspective it's a breaking change, but since we always passed `undefined` and removing the type means it's still `undefined` so it's practically speaking the same.
- Clarify docs on `generateUrl`, `getFileIdFromRequest`, and `namingFunction`. This is a bit awkward and it seems we didn't think about this enough in #515. Let me explain:

I thought `generateUrl` supersedes `namingFunction`, as instead of only generating an ID we now do the entire URL. Except `generateUrl` only changes the returned `Location` header, while we used to have the principle of upload ID mapping to the file name in storage. So if you return an entirely different ID in `generateUrl`, the default of `namingFunction` is the name written to storage. If you define `namingFunction`, then that becomes the `id` argument in `generateUrl`. This is a bit of an unobvious and brittle relationship. 

Ideally, `namingFunction` is deprecated and the ID of `generateUrl` is used for creation in stores too. But the problem is `generateUrl` already expects the upload ID as the second argument.

Since we likely can't solve this without a breaking change I made this PR to improve the docs at least. 

How do we make this better?

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -78,17 +78,47 @@ Allow `Forwarded`, `X-Forwarded-Proto`, and `X-Forwarded-Host` headers to overri
 Additional headers sent in `Access-Control-Allow-Headers` (`string[]`).
 
 #### `options.generateUrl`
-Control how the upload url is generated (`(req, { proto, host, baseUrl, path, id }) => string)`)
+
+Control how the upload URL is generated (`(req, { proto, host, path, id }) => string)`)
+
+This only changes the upload URL (`Location` header).
+If you also want to change the file name in storage use `namingFunction`.
+Returning `prefix-1234` in `namingFunction` means the `id` argument in `generateUrl` is `prefix-1234`.
+
+```js
+function generateUrl(req, {proto, host, path, id}) {
+  const prefix = getPrefixForUser(req) // your custom logic
+  return `${proto}://${host}${path}/${prefix}/${id}?query=param`
+},
+```
+
+> [!NOTE] > `@tus/server` expects everything after the last `/` to be the upload id.
+> If you change that you have to use `getFileIdFromRequest` as well.
 
 #### `options.getFileIdFromRequest`
+
 Control how the Upload-ID is extracted from the request (`(req) => string | void`)
 
 #### `options.namingFunction`
 
 Control how you want to name files (`(req) => string`)
 
+In `@tus/server`, the upload ID in the URL is the same as the file name.
+This means using a custom `namingFunction` will return a different `Location` header for uploading
+and result in a different file name in storage.
+
 It is important to make these unique to prevent data loss. Only use it if you need to.
 Default uses `crypto.randomBytes(16).toString('hex')`.
+
+```js
+function namingFunction(req) {
+  const prefix = getPrefixForUser(req) // your custom logic
+  return `${prefix}-${crypto.randomBytes(16).toString('hex')}`
+},
+```
+
+> [!CAUTION]
+> You can not use slashes (`/`) in your name.
 
 #### `disableTerminationForFinishedUploads`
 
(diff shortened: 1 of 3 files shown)


packages/server/README.md
```diff
@@ -78,18 +78,49 @@ Allow `Forwarded`, `X-Forwarded-Proto`, and `X-Forwarded-Host` headers to overri
...
+}
+```
+
+> [!NOTE]
+> `@tus/server` expects everything after the last `/` to be the upload id.
+> If you change that you have to use `getFileIdFromRequest` as well.
 
 #### `options.getFileIdFromRequest`
+
 Control how the Upload-ID is extracted from the request (`(req) => string | void`)
```
Acconut, line 101:
```suggestion
Control how the Upload-ID is extracted from the request (`(req) => string | void`)
By default, it expects everything in the path after the last `/` to be the upload id.
```

packages/server/README.md
```diff
@@ -78,18 +78,49 @@ Allow `Forwarded`, `X-Forwarded-Proto`, and `X-Forwarded-Host` headers to overri
 Additional headers sent in `Access-Control-Allow-Headers` (`string[]`).
 
 #### `options.generateUrl`
-Control how the upload url is generated (`(req, { proto, host, baseUrl, path, id }) => string)`)
+
+Control how the upload URL is generated (`(req, { proto, host, path, id }) => string)`)
+
+This only changes the upload URL (`Location` header).
```
Acconut, line 84: Can we add an exemplary use case when generateUrl is useful? For example, to rewrite the URL so that it fits the format that is expect by a proxy for incoming requests?

packages/server/README.md
```diff
@@ -78,18 +78,49 @@ Allow `Forwarded`, `X-Forwarded-Proto`, and `X-Forwarded-Host` headers to overri
...
+
+```js
+function generateUrl(req, {proto, host, path, id}) {
+  const prefix = getPrefixForUser(req) // your custom logic
+  return `${proto}://${host}${path}/${prefix}/${id}?query=param`
+}
+```
+
+> [!NOTE]
+> `@tus/server` expects everything after the last `/` to be the upload id.
```
Acconut, line 96:
```suggestion
> `@tus/server` expects everything in the path after the last `/` to be the upload id.
```

This makes it clear that the query string is not interpreted as part of the upload ID. Or is this the case?

Acconut (commented):
Great improvement! After these changes, the two options don't seem odd to me anymore. `namingFunction` customizes the upload ID and file name, while `generateURL` and `getFileIdFromRequest` (which should maybe have been called `getUploadIdFromRequest`?) embed or extract the upload ID from the URL.

packages/server/README.md
Murderlon: I did add an example for adding a prefix based on the user from the request. Unless we know a common format change for a proxy, I think the result would also just be another psuedo-code example with some non existing functions to call.

packages/server/README.md
Murderlon: Good one.

https://github.com/tus/tus-node-server/blob/421edee819f579ceefe85595ce9da23f240cc5bb/packages/server/test/BaseHandler.test.ts#L73-L91

packages/server/README.md
Acconut: I wasn't thinking about code here, but an example situation explained in prosa where `generateUrl` is necessary or helpful. This makes it easier for users to understand when they would want to change the upload URL.

Murderlon:
To simplify the code for custom implementations, we could also pass a default id to `namingFunction` and the default extracted id to `getFileIdFromRequest`. Then it becomes a lot cleaner:

```diff
const path = '/files'
const server = new Server({
  path,
  datastore: new FileStore({directory: './test/output'}),
-  namingFunction(req) {
+ namingFunction(req, id) {
-    const id = crypto.randomBytes(16).toString('hex')
    const folder = getFolderForUser(req) // your custom logic
    return `users/${folder}/${id}`
  },
  generateUrl(req, {proto, host, path, id}) {
    id = Buffer.from(id, 'utf-8').toString('base64url')
    return `${proto}://${host}${path}/${id}`
  },
-  getFileIdFromRequest(req) {
+ getFileIdFromRequest(req, id) {
-    const reExtractFileID = /([^/]+)\/?$/
-    const match = reExtractFileID.exec(req.url as string)
-
-   if (!match || path.includes(match[1])) {
-      return
-   }
-
-    return Buffer.from(match[1], 'base64url').toString('utf-8')
+    return Buffer.from(id, 'base64url').toString('utf-8')
  },
})
```

fenos:
@Murderlon not sure i like the idea of passing a default id in the `namingFunction` - it feels like wasted computation for users who really don't need the `defaultId`

Murderlon:
Fair enough, I was mostly thinking out loud. We can keep things as is for now. `uri` probably isn't different from `req.url` either.

fenos (approved):
Looks great!

PR #552, opened 2024-01-15
@s3/store: allow providing custom cache implementation

fenos:
Currently, S3 metadata is always cached in memory using a `Map`.
This increases performance, but it becomes problematic when running multiple instances of tus-node-server.

Since the cache is not shared between the instances, when a file is deleted, other instances might still have the cached record and there is no way to invalidate it.

This PR  allows for 3rd party cache implementations such as with Redis, in order to share the cache within a cluster of tus-node-server.

It also provides a `NullCache` in case the user wants to disable the cache


Murderlon (commented):
I think it's a great idea but I'm worried we're going to keep repeating ourselves in every store. We already have `FileConfigstore`, `MemoryConfigStore`, and `RedisConfigstore`. These caches are practically the same.

Some options:
1. Put config stores in `@tus/server`, reuse them in stores. We could leave out `@tus/gcs-store` for now.
   - Optional: rename them to caches
2. Make it a server option and reuse it in all stores. Bit more involved as we'd have to write a fallback for `file-store` to not make it a breaking change and it's perhaps expected that it would work in `gcs-store` too.

What do you think?

fenos:
Hi @Murderlon currently it seems that `FileConfigstore, MemoryConfigStore, and RedisConfigstore` are different than the S3 metadata cache?

From my understanding, the S3Store doesn't need a `ConfigStore` because it stores the metadata in a .info file.
However, the implementation tries to minimize the calls to S3 to fetch the .info file, essentially a cache.

Where the file store is using any of the above-mentioned stores to store the `.info` file content. Not as a cache.

fenos:
@Murderlon I gave it a second thought and it makes a lot of sense what you have said above.
I went with option 1 and copied over the `configstores` to @tus/server - we can deprecate the configstores in the filestore in a later release

Murderlon (commented):
Great 👍 

Since we're keeping the config stores in file store for now and the addition of them in `@tus/server` is new, shall we rename them to "caches" (`FileCache`)? The "configstore" is a remnant from the past when the `configstore` package was used which placed the info files in `~/.config`, which was very odd.

fenos:
@Murderlon interesting,

This is a tricky situation for naming since these "stores" are not really caches, they can be used as such but they can also be intended to be used as a persistent layer, especially for the `file-store`.

I think `configstore` is still a better generic name atm.
we could call them `stores` or `metadatastores` with the former being more confusing than ever 😄 

what do you think?

fenos:
I changed them to be `kvstores` ex:

- FileKvStore
- RedisKvStore
- MemoryKvStore

Murderlon (approved):
Now that I think of it naming is perhaps a bit tricky. They all don't feel completely right but "kv" seems good enough.

Murderlon:
I'm adding some new docs now.

fenos:
Thanks a lot @Murderlon 🙏

PR #553, opened 2024-01-16
@s3/store: Allow disabling tagging on expiration extension

### LINKED ISSUE #538: @tus/s3-store: add option to disable object tags
I was using uppy with tusd and the S3 datastore -- which works great. But my project is pretty much 90% JS so I wanted to reuse some stuff like my JS database client. This led me to give tus-node-server a try. With @tus/file-store it works fine, but with @tus/s3-store uppy fails to upload the file with the following error: `Something went wrong with that request Invalid URL`

I'm using pretty much the exact sample from the s3Store package's readme:

```
const {Server, EVENTS} = require("@tus/server");
const {S3Store} = require("@tus/s3-store");

// this works correctly and logs the same correct values that work perfectly with tusd
console.log(process.env.S3_BUCKET, process.env.S3_REGION, process.env.S3_ENDPOINT, process.env.S3_ACCESS_KEY_ID, process.env.S3_SECRET_ACCESS_KEY);

const s3Store = new S3Store({
  s3ClientConfig: {
    bucket: process.env.S3_BUCKET,
    region: process.env.S3_REGION,
    endpoint: process.env.S3_ENDPOINT,
    credentials: {
      accessKeyId: process.env.S3_ACCESS_KEY_ID,
      secretAccessKey: process.env.S3_SECRET_ACCESS_KEY,
    },
  },
});

const tusServer = new Server({
  path: "/files",
  datastore: s3Store,
});

console.log("Listening on :3000");
tusServer.listen({host: '0.0.0.0', port: 3000});
```

My uppy config is: 

```
new Uppy({meta: {shop: shop}}).use(Tus, {
  endpoint: 'https://node-tus-xxxxxxxxxx.up.railway.app/files/',
}),
```

fenos:
This PR fixes #538 introduces a new option `useTags` which will allow disabling tagging objects for S3 providers who don't support them.

Tags will be now used only if the expiration extension is used.

### DIFF
--- a/packages/s3-store/README.md
+++ b/packages/s3-store/README.md
@@ -71,6 +71,17 @@ Options to pass to the AWS S3 SDK.
 Checkout the [`S3ClientConfig`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html)
 docs for the supported options. You need to at least set the `region`, `bucket` name, and your preferred method of authentication.
 
+
+#### `options.expirationPeriodInMilliseconds`
+
+Enables the expiration extension and sets the expiration period of an upload url in milliseconds.
+Once the expiration period has passed, the upload url will return a 410 Gone status code.
+
+#### `options.useTags`
+
+Some S3 providers don't support tagging objects.
+If you are using certain features like the expiration extension and your provider doesn't support tagging, you can set this option to `false` to disable tagging.
+
 ## Extensions
 
 The tus protocol supports optional [extensions][]. Below is a table of the supported extensions in `@tus/s3-store`.
--- a/packages/s3-store/index.ts
+++ b/packages/s3-store/index.ts
@@ -22,6 +22,7 @@ type Options = {
   // The server calculates the optimal part size, which takes this size into account,
   // but may increase it to not exceed the S3 10K parts limit.
   partSize?: number
+  useTags?: boolean
   expirationPeriodInMilliseconds?: number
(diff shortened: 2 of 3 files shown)


rougsig:
Will expiration extension work if tagging disabled?

If not, may be it's better to disable extension from options?

packages/s3-store/README.md
```diff
@@ -71,6 +71,17 @@ Options to pass to the AWS S3 SDK.
...
+
+#### `options.expirationPeriodInMilliseconds`
+
+Enables the expiration extension and sets the expiration period of an upload url in milliseconds.
+Once the expiration period has passed, the upload url will return a 410 Gone status code.
+
+#### `options.expirationExtensionDisableTags`
+
+Some S3 providers don't support tagging objects.
+If you are using the expiration extension and your provider doesn't support tagging, you can set this option to `true` to disable tagging.
```
Murderlon, line 83: How about renaming this to `useTags` which defaults to `true`. Reasons:

- "If you are using the expiration extension" is not really a thing because it's always there available
- "Disable" with "true" is a bit odd compared to "enabled" with "true".
- The problem is tags in general, not just with the expiration extension, so an option to disable all tags makes sense to me.

Murderlon (commented):
Thanks for picking this up 👌

packages/s3-store/README.md
fenos: Sure, happy to rename it

fenos: Done, renamed to `useTags`

Murderlon (approved):
💯

PR #558, opened 2024-01-17
@tus/server: Metadata on namingFunction + propagate error

fenos:
This small PR simply allows error propagation in the namingFunction.

Currently, if an error is thrown from the `namingFunction` the error is swallowed and a 500 error code is returned.
This behavior is not ideal, in my use case I rely on the `metadata` in order to create a valid uploadID.

For this reason the `metadata` is validated in the naming function. If some of the fields don't pass the validation instead of returning a 400 status code, the application returns a non-ideal 500

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -104,7 +104,7 @@ By default, it expects everything in the path after the last `/` to be the uploa
 
 #### `options.namingFunction`
 
-Control how you want to name files (`(req) => string`)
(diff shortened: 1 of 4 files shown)


Murderlon:
Sounds good 👍 

We can add the `metadata` argument in this PR as well

fenos:
Done 🎉

Murderlon (approved):

Murderlon:
I'll do a release tomorrow.

Issue #505, opened 2023-10-23
@tus/s3-store: Issues with StreamSplitter


mitjap:
Current implementation of stream splitter has some issues/drawbacks. I think these should be addressed in near future. 

Parallel nature of this class can cause file corruption when upload of any part fails (except the one which is last in that moment). 
If upload of any part fails when less than 5MB was transferred to S3 then this upload will fail when finalizing multipart upload.

We could solve this by acknowledging when part is uploaded and consider only these parts to be valid. When calculating current offset (on HEAD request) we need to consider only first consecutive valid parts.

No back-pressure support. When TUS server is running on local network, uploads will be cached to disk without any consideration of actual upload speed to S3.

There is no need to constantly split to smaller chunks. Server should cache only first 5MB of data to determine if it is safe to upload as a proper "part" and if it is, then it should stream directly to S3, but only up to 5GB for each part.

Possible implementation of stream splitter would work like this. On PATCH request an incomplete part would be immediately downloaded to temporary file. Current PATCH request would append to this temporary file. If request is aborted/completed before 5MB threshold is reached that this data is uploaded as incomplete part. If 5MB threshold is reached then we should stop buffering to file and upload temporary file to S3. When this is done we would redirect PATCH request directly to S3. Once current part size reaches 5GB process is repeated without downloading of incomplete part.

fenos:
> Parallel nature of this class can cause file corruption when upload of any part fails (except the one which is last in that moment).
> If upload of any part fails when less than 5MB was transferred to S3 then this upload will fail when finalizing multipart upload.

This is a valid point, I believe that tusd suffers from the same issue, see 

- UploadParts https://github.com/tus/tusd/blob/main/pkg/s3store/s3store.go#L501-L547
- ListAllParts https://github.com/tus/tusd/blob/main/pkg/s3store/s3store.go#L1028-L1059

CC: @Acconut 

> We could solve this by acknowledging when part is uploaded and consider only these parts to be valid. When calculating current offset (on HEAD request) we need to consider only first consecutive valid parts.

This could be nice! yes. 
CC: @Acconut 

> No back-pressure support. When TUS server is running on local network, uploads will be cached to disk without any consideration of actual upload speed to S3.

This has been implemented #561 

> There is no need to constantly split to smaller chunks. Server should cache only first 5MB of data to determine if it is safe to upload as a proper "part" and if it is, then it should stream directly to S3, but only up to 5GB for each part.

[...]

PR #561, opened 2024-01-19
@s3/store: add backpressure + download incomplete part first

### LINKED ISSUE #501: S3Store: Deferred length uploads fail when only 1 incomplete part is uploaded
When uploading a readable stream (deferred length), when the overall upload is smaller than the [minPartSize](https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L73C10-L73C21), then the upload remains as an `incomplete part`, never establishing an actual `part` for the S3 multipart upload to complete.

## Reproducible Steps

1. Upload via readable stream with deferred lengths
2. Upload a very small set of chunks that are collectively smaller than the [minPartSize](https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L73C10-L73C21)

## Expected behavior
The file upload completes

## Actual behavior
`500` - [finishMultipartUpload](https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L501-L502) fails because there are no completed `parts` resolved from [client.listParts](https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L403).

As a result, S3 contains no `Parts` in its response field, resulting in an `undefined` error when [finishMultipartUpload](https://github.com/tus/tus-node-server/blob/0f0e5d390f42cc8c61f815cb53db7404c96e5600/packages/s3-store/index.ts#L362-L367) attempts `parts.map`

Fix - https://github.com/tus/tus-node-server/pull/502

fenos:
This PR improves the handling of incomplete parts and adds backpressure by limiting the number of concurrent uploads/files on the disk that are being created.

## Incomplete Parts

The incomplete part logic was somewhat not always working as intended.

An issue I came across with the old implementation is that when i upload a file using chunked upload from the client, and the chunk is bigger than the `preferred` size on the server, at times the offset returned wasn't calculated properly resulting in the subsequent request to return 409 (Offset-Conflict) very much related to  #501

This PR fixes this issue, by always downloading the incomplete part first and prepending it to the readstream.
By using this approach the offset is now calculated properly and also fixes #501 (see test) 

## Backpressure

Added support for backpressure. The way it works is by using a semaphore on the StreamSplitter.
Only X number of chunks on disk can be created at any one time (default 30) and the semaphore is released upon finishing the part upload, meaning that the backpressure is handled by how fast S3 can upload parts.

In the scenario when the limit of concurrent chunks has been reached, the default node mechanism of backpressure will kick in. This means that when the highwater mark on the readstream buffer is full,  the stream will be paused until the flow of data continues and the semaphore has capacity.

This aligns with the issue #505


packages/server/src/handlers/BaseHandler.ts
```diff
@@ -151,9 +151,11 @@ export class BaseHandler extends EventEmitter {
         reject(err.name === 'AbortError' ? ERRORS.ABORTED : err)
       })
 
-      req.on('error', (err) => {
+      req.on('error', () => {
         if (!proxy.closed) {
-          proxy.destroy(err)
+          // we end the stream gracefully here so that we can upload the remaining bytes to the store
+          // as an incompletePart
+          proxy.end()
```
fenos, line 158: A very small but nice change here.
When an upload request is canceled, Instead of calling `proxy.destroy()` we can call `proxy.end()` to gracefully close the passthrough stream, so that we can flush the bytes of a chunk as an incomplete part and upload it to S3.

This will result in starting exactly where we left off as opposed to starting from the last successful chunk uploaded.

packages/s3-store/index.ts
```diff
@@ -25,6 +28,7 @@ type Options = {
   // but may increase it to not exceed the S3 10K parts limit.
   partSize?: number
   useTags?: boolean
+  maxConcurrentPartUploads?: number
```
Murderlon, line 31: Do we need to expose this as an option? I think this falls under the category of we as upload experts should determine what is best for users for this low-level option.

packages/s3-store/index.ts
```diff
@@ -741,4 +809,26 @@ export class S3Store extends DataStore {
...
+      fs.lstat(filePath, (error) => {
+        if (!error) {
+          return resolve(true)
+        }
+        if (error.code === 'ENOENT') {
+          return resolve(false)
+        }
+        reject(error)
+      })
+    })
```
Murderlon, line 827: Would be nice to start to transition to `fs/promises` where possible, especially new additions.

Murderlon (commented):
A great addition!

Might need to go over it one more time but some initial feedback here.

packages/s3-store/index.ts
fenos: It all depends on the upload bandwidth of where the server is running, there is no magic value to be fair.
At the moment we are simply limiting how many parallel chunks and uploads can be processed at any point in time. 

For example, if you are running the server in a local network with a local minio, this value could be much higher since the upload throughput is higher.

fenos: similarly, you run the uploads within the AWS platform (ECS/ECS, etc..) against S3, the upload would be faster than if you upload it from a server in a different provider. Meaning the concurrentPartUploads could increased.

My default value of max 30 concurrent chunks with a preferred size of 8MB means that the upload bandwidth will be around 240MB/s - the file system will match that with backpressure. 

Without backpressure, the file system could write parts at a speed of 1GB/s but upload at much slower rate

packages/s3-store/index.ts
fenos: Yes, I just found it less verbose using the callback instead of `try { } catch {} ` for this instance.
But happy to migrate to fs/promises if preferred

packages/server/src/models/StreamSplitter.ts
```diff
@@ -117,13 +125,34 @@ export class StreamSplitter extends stream.Writable {
     this.part += 1
   }
 
+  async emitEvent<T>(name: string, payload: T) {
+    if (this.asyncEvents) {
```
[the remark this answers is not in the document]
fenos: a bit less canonical but working with async events from the event emitter is also not that great 😄

packages/s3-store/index.ts
fenos: Done

packages/server/src/models/StreamSplitter.ts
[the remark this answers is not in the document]
fenos: Oh! 
I just learned a neat wait to have async compatible events!

instead of calling `emit('name', payload)`

we implement our own `emitAsync` function:

```ts
  async emitEvent<T>(name: string, payload: T) {
    const listeners = this.listeners(name)
    for await (const listener of listeners) {
      await listener(payload)
    }
  }
```

and voila'

packages/s3-store/index.ts
Murderlon: That would be a bit too naive to not expose it. Perhaps we want to write some docs about the considerations, as I do think tus-node-server is also popular under front-end/full-stack devs with less awareness of this. 

how about something like this in the s3-store readme:

### Example: using `partSize` and `maxConcurrentParts` efficiently

These settings depend on the machine's processing power, memory, network bandwidth, and disk I/O capabilities.
Below are some recommendations to take into account but exact settings depend on your setup.

- **Uploading within the AWS platform to S3**. Given the high-speed environment and assuming robust server resources both values can be set higher than the default.
- **Uploading from outside the AWS platform to S3**.  These are variable-speed environments with more network fluctuations. The default `partSize` of 8 MiB and default `maxConcurrentParts` of 30 could work well for this (or lower on a resource-limited machine).
- **Using Local MinIO** (Local Network or On-Premises). Typically high-speed and low-latency but this should be balanced against the capabilities of your local server infrastructure. Values similar to or higher than those used within AWS might be suitable.

In general, the tradeoff is reliability vs. speed. A bigger `partSize` requires less roundtrips but is more prone to failure in less stable connections. Higher concurrency (`maxConcurrentParts`) increases CPU and memory usage.

fenos:
@Murderlon I've updated the Readme and added a detailed documentation on the `maxConcurrentPartUploads` flag.

I've also increased the value to 60 instead of 30 (~480MB), as this aligns better with tusd default.
Tusd uses 50MB `partSize` and 10 `maxConcurrentPartUploads` totaling at 500MB.

Let me know if there is anything else, this change will be awesome to have in

packages/s3-store/README.md
```diff
@@ -91,6 +91,22 @@ you need to provide a cache implementation that is shared between all instances
 
 See the exported [KV stores][kvstores] from `@tus/server` for more information.
 
+### `options.maxConcurrentPartUploads`
```
Murderlon, line 94:
```suggestion
#### `options.maxConcurrentPartUploads`
```

packages/s3-store/README.md
```diff
@@ -91,6 +91,22 @@ you need to provide a cache implementation that is shared between all instances
...
+**Considerations:**
+The ideal value for maxConcurrentPartUploads varies based on your `partSize` and the upload bandwidth to your S3 bucket. A larger partSize means less overall upload bandwidth available for other concurrent uploads.
+
+- **Lowering the Value**: Reducing maxConcurrentPartUploads decreases the number of simultaneous upload requests to S3. This can be beneficial for conserving memory, CPU, and disk I/O resources, especially in environments with limited system resources or where the upload speed it low or the part size is large.
+
+
+- **Increasing the Value**: A higher value potentially enhances the data transfer rate to the server, but at the cost of increased resource usage (memory, CPU, and disk I/O). This can be advantageous when the goal is to maximize throughput, and sufficient system resources are available.
+
+
+- **Bandwidth Considerations**: It's important to note that if your upload bandwidth to S3 is a limiting factor, increasing maxConcurrentPartUploads won’t lead to higher throughput. Instead, it will result in additional resource consumption without proportional gains in transfer speed.
```
Murderlon, line 108:
```suggestion
The ideal value for `maxConcurrentPartUploads` varies based on your `partSize` and the upload bandwidth to your S3 bucket. A larger partSize means less overall upload bandwidth available for other concurrent uploads.

- **Lowering the Value**: Reducing `maxConcurrentPartUploads` decreases the number of simultaneous upload requests to S3. This can be beneficial for conserving memory, CPU, and disk I/O resources, especially in environments with limited system resources or where the upload speed it low or the part size is large.


- **Increasing the Value**: A higher value potentially enhances the data transfer rate to the server, but at the cost of increased resource usage (memory, CPU, and disk I/O). This can be advantageous when the goal is to maximize throughput, and sufficient system resources are available.


- **Bandwidth Considerations**: It's important to note that if your upload bandwidth to S3 is a limiting factor, increasing `maxConcurrentPartUploads` won’t lead to higher throughput. Instead, it will result in additional resource consumption without proportional gains in transfer speed.
```

Murderlon (approved):
LGTM 👌 

I think I would like a quick review from @Acconut on this before merging.

packages/s3-store/README.md
```diff
@@ -91,6 +91,22 @@ you need to provide a cache implementation that is shared between all instances
 
 See the exported [KV stores][kvstores] from `@tus/server` for more information.
 
+#### `options.maxConcurrentPartUploads`
+
+This setting determines the maximum number of simultaneous part uploads to an S3 storage service.
```
Acconut, line 96: It might be helpful to state in this document what a part actually is. Maybe a brief explanation saying that this module uses S3 multipart uploads, which splits a file into multiple equally-sized parts (independent of the PATCH requests). With this context, it's easier to understand what concurrent part uploads actually means.

packages/s3-store/README.md
```diff
@@ -91,6 +91,22 @@ you need to provide a cache implementation that is shared between all instances
...
+The default value is 60. This default is chosen in conjunction with the typical partSize of 8MiB, aiming for an effective transfer rate of approximately 480MiB/s.
```
Acconut, line 97: The calculation seems assume that one part is uploaded in 1s. Might be worth mentioning. Also transfer speeds are commonly given in Mbit/s not MiB/s. That would be 3.84 Gbit/s, which seems very high. I am not sure if the transfer speeds to S3 reach that level.

packages/s3-store/index.ts
```diff
@@ -42,6 +46,45 @@ function calcOffsetFromParts(parts?: Array<AWS.Part>) {
   return parts && parts.length > 0 ? parts.reduce((a, b) => a + b.Size, 0) : 0
 }
 
+function concatenateStreams(streams: Readable[]): Readable {
```
Acconut, line 49: Is it possible to use an existing module for that, e.g. https://www.npmjs.com/package/multistream? Then we would not have to maintain this logic.

packages/s3-store/index.ts
```diff
@@ -82,6 +125,7 @@ export class S3Store extends DataStore {
   private preferredPartSize: number
   private expirationPeriodInMilliseconds = 0
   private useTags = true
+  private semaphore: Semaphore
```
Acconut, line 128:
```suggestion
  private partUploadSemaphore: Semaphore
```

A good name helps understand what the semaphore actually controls.

packages/s3-store/index.ts
```diff
@@ -233,6 +278,61 @@ export class S3Store extends DataStore {
...
+
+    if (!incompletePart) {
+      return
+    }
+    const filePath = await this.uniqueTmpFileName('tus-s3-incomplete-part-')
+
+    try {
+      let incompletePartSize = 0
+
+      const byteCounterTransform = new stream.Transform({
```
Acconut, line 292: Using a transformer to obtain the incomplete part's size appear a bit complex to me. Two alternatives are either using the length that is returned in the `GetObject` call to S3 (AFAIK there should be a length in there) or reading the file size after the part has been written to disk. Then we can remove this additional stream.

Acconut: I just saw that this code was copied from somewhere else. Feel free to ignore it then :)

packages/s3-store/index.ts
```diff
@@ -484,9 +536,11 @@ export class S3Store extends DataStore {
     else if (size <= this.preferredPartSize * this.maxMultipartParts) {
       optimalPartSize = this.preferredPartSize
       // The upload is too big for the preferred size.
-      // We devide the size with the max amount of parts and round it up.
+      // We divide the size with the max amount of parts and round it up.
```
Acconut, line 539: There is no rounding up here happening. The comment likely needs to be updated.

packages/s3-store/index.ts
```diff
@@ -484,9 +536,11 @@ export class S3Store extends DataStore {
...
+    } else if (size % this.maxMultipartParts === 0) {
+      optimalPartSize = size / this.maxMultipartParts
     } else {
-      optimalPartSize = Math.ceil(size / this.maxMultipartParts)
+      optimalPartSize = size / this.maxMultipartParts + 1
```
Acconut, line 543: Why was the `Math.ceil` removed? If we reach this conditional block, we know that `size / this.maxMultipartParts` is not an integer, so rounding it would be sense to get rid of the fractional part.

packages/s3-store/index.ts
```diff
@@ -233,6 +278,61 @@ export class S3Store extends DataStore {
...
+      })
+
+      // write to temporary file
+      await streamProm.pipeline(
+        incompletePart,
+        byteCounterTransform,
+        fs.createWriteStream(filePath)
+      )
+
+      const createReadStream = (options: {cleanUpOnEnd: boolean}) => {
```
Acconut, line 306: `cleanUpOnEnd` seems to always be true. Do we need this option then?

packages/s3-store/index.ts
```diff
@@ -741,4 +799,29 @@ export class S3Store extends DataStore {
 
     return deleted
   }
+
+  private async uniqueTmpFileName(template: string): Promise<string> {
```
Acconut, line 803: Is there also a public package that we could reuse here?

packages/server/src/handlers/BaseHandler.ts
Acconut: Nice!

packages/server/src/models/Semaphore.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+}
+
+interface Deferred {
+  resolve?(permit: Permit): void
+  promise?: Promise<Permit>
+}
+
+export class Semaphore {
```
Acconut, line 27: Nice implementation, but I am wondering again whether we can reuse existing module for this.

packages/server/src/models/StreamSplitter.ts
```diff
@@ -106,7 +108,7 @@ export class StreamSplitter extends stream.Writable {
 
     await this.fileHandle.close()
 
-    this.emit('chunkFinished', {
+    await this.emitEvent('chunkFinished', {
```
Acconut, line 111: Can you explain why this change was necessary?

packages/server/src/models/StreamSplitter.ts
```diff
@@ -117,13 +119,23 @@ export class StreamSplitter extends stream.Writable {
     this.part += 1
   }
 
+  async emitEvent<T>(name: string, payload: T) {
```
Acconut, line 122: This function should be named differently because it doesn't emit events in an asynchronous fashion like we are used from EventEmitter. Instead, it sequentially invokes callbacks, which is different.

Acconut (changes requested):
Amazing work, @fenos!

packages/server/src/models/StreamSplitter.ts
fenos: the standard EventEmitter emits events **synchronously** this function does the same thing but it allows using async/await to have a similar behavior to the standard EventEmitter

packages/server/src/models/StreamSplitter.ts
fenos: to keep the events interface consistent

packages/server/src/models/Semaphore.ts
fenos: Yea, I thought that minimizing dependencies for these utility classes could be ok

packages/s3-store/index.ts
fenos: Ah nice! sure happy to use `multistream` in this case

packages/s3-store/index.ts
fenos: i think this is such a basic functionality that a package might be not necessary, wdyt?

packages/server/src/models/StreamSplitter.ts
fenos: In essence the standard EventEmitter loops sequentially all events as my implementation and the handler blocks in a synchronous fashion, however, if you do async operations it won't await by default

Acconut:
Regarding replacing internal code with dependencies, I would wait for a comment from @Murderlon. Maybe his position is different than mine :)

packages/server/src/models/StreamSplitter.ts
Acconut: You are right, thanks for correct me! Then the name is fine.

packages/s3-store/index.ts
Murderlon: If this single function does the job, I personally don't see the reason to add more dependencies, unless there are other benefits.

Murderlon:
> Regarding replacing internal code with dependencies, I would wait for a comment from @Murderlon. Maybe his position is different than mine :)

I'd personally would avoid adding a dependency if it's not much code, we likely rarely/never have to change it, and we don't foresee complex side-effects. So I don't think we need a package to concat streams or unique tmp files. For the semaphore I don't have a strong opinion, in this case it's a well tested copy paste from Shopify so from a trust perspective I'd say it's equal to a package. But if there is a similar well tested package with a small footprint I would be okay with that too. I'm okay with either.

packages/s3-store/index.ts
fenos: I based this change on this:
https://github.com/tus/tusd/blob/main/pkg/s3store/s3store.go#L1176

There is no rounding as far as i can see

packages/s3-store/index.ts
fenos: Yhea, this option is more for awareness, that at the end of the stream, the file is removed.

It could hide the detail in which the file is being removed to the next developer looking at the code if we simply call: `createReadStream()` without the option.

Happy to remove it, but it seems clearer with it

packages/s3-store/index.ts
Murderlon: Go automatically does rounding :)

Acconut: To add some more explanation to @Murderlon's comment: The code from tusd uses integers. Integer division is different than floating point division in that the fractional is ignored. For example:

```
14 / 5 == 2 // integer division
14.0 / 5 == 2.8 // float divison
```

That's why you want `Math.floor` when trying to replicate the Go logic, or `Math.ceil` when you want to replicate the `integer division + 1` from tusd.

Acconut:
Alright, if you prefer less dependencies, we can go with that route as well. That's probably a topic where personal preferences differ.

packages/s3-store/index.ts
fenos: Ah yes, very good point! I totally missed that the data types were ints and not float. Will bring the rounding back :)

Murderlon:
I just saw that [`@shopify/semaphore`](https://www.npmjs.com/package/@shopify/semaphore) is available as a package. In that case I agree with @Acconut that a dependency is better.

packages/s3-store/README.md
fenos: Traffic between Amazon EC2 and Amazon S3 uses up to 100 Gbps of bandwidth to Amazon Virtual Private Cloud (Amazon VPC) endpoints and public IP addresses in the same AWS Region.

packages/s3-store/index.ts
```diff
@@ -25,6 +28,7 @@ type Options = {
   // but may increase it to not exceed the S3 10K parts limit.
   partSize?: number
   useTags?: boolean
+  maxConcurrentPartUploads?: number | Semaphore
```
Murderlon, line 31: It's not documented that this can also be a `Semaphore`. Any reason against keeping this simple and only doing a `number`, which we always pass to `Semaphore`?

Murderlon (commented):
Changes look good 👍 One Q

packages/s3-store/index.ts
fenos: I thought that in case a server runs multiple instances of TusServer they might want to share a single Semaphore (which I think in the future we should also allow passing an S3Client too)

Will document now

Murderlon: Is that something Supabase already needs? I'm also okay with keeping it simple for now until the need arrises, then we would add a new option which would also be less odd than abusing `maxConcurrentPartUploads` a bit.

fenos: We don't need it right now, but probably in the near future.
Happy to revisit this later

fenos: Reverted this

fenos:
@Murderlon all ready

Murderlon (approved):
Thanks!

PR #563, opened 2024-01-21
@tus/server: add possibility to have an async namingFunction

mattiasrunge:
Supporting an async namingFunction enables the check of existing files to ensure there are no collisions.

### DIFF
--- a/packages/server/src/handlers/PostHandler.ts
+++ b/packages/server/src/handlers/PostHandler.ts
@@ -65,7 +65,7 @@ export class PostHandler extends BaseHandler {
 
     let id
     try {
-      id = this.options.namingFunction(req, metadata)
+      id = await Promise.resolve(this.options.namingFunction(req, metadata))
     } catch (error) {
       log('create: check your `namingFunction`. Error', error)
       throw error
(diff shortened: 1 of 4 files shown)


packages/server/src/handlers/PostHandler.ts
```diff
@@ -65,7 +65,7 @@ export class PostHandler extends BaseHandler {
 
     let id
     try {
-      id = this.options.namingFunction(req, metadata)
+      id = await Promise.resolve(this.options.namingFunction(req, metadata))
```
fenos, line 68: you don't technically need the `Promise.resolve` wrapper, you can simply await it, makes it cleaner

fenos:
Awesome! Thanks a lot for the PR! looks like having the namingFunction async has its use cases!

packages/server/src/handlers/PostHandler.ts
mattiasrunge: You are right of course, I did actually did not know that so now I have learned something as well, thanks! :)

mattiasrunge:
Not sure why the tests failed this time, the change should not affect S3 which the test seemed related to, is this something intermittent? Or have I missed something?

Murderlon (approved):

Murderlon:
Thanks!

PR #567, opened 2024-01-29
Introduce @tus/utils to simplify building/publishing

Murderlon:
**Problems**:
- In all packages we have `@tus/server` in `peerDependencies` but also in `devDependencies` (for tests).
- `peerDependencies` changes aren't noticed by package install unless it's a major. This is why `changeset` (our versioning and release tool) _always_ bumps all stores to a major for any kind of release to `server`. See #566 
- Turbo can't cache correctly when running tests. For instance, `file-store` depends on `server` but `server` tests depend on `file-store`. We can't create a circular dependency so we omit `file-store` as dev but then changes aren't picked up. 

**Solution**:
- Introduce `@tus/utils`, have it in `dependencies` everywhere. All things that are shared between all packages are in here. Remove `@tus/server` from `peerDependencies` everywhere.
- No breaking changes in exports.

Now we have correct dependency graphs for `turbo` and we can publish more correctly and easily with `changesets`.

This is arguably a breaking change, but I'd argue it isn't. It only is if you update one package but not the other. However that can happen all time the time in monorepo setups like this. Whether it's a bug fix that is coordinated between packages, or a feature that needs detecting in a store, you can't expect all changes to work if you don't upgrade packages together. In Uppy, our policy is it's not a breaking change if we don't change exports and no public methods and properties a are changed.


Acconut:
I don't understand how @tus/utils helps with the `peerDependencies` issue. Can you elaborate on this? Or is not addressed by this PR?

Murderlon:
As seen in the diff, there are no peer dependencies anymore and thus no problems anymore associated with them. All packages now have `@tus/utils` in `dependencies`.

PR #568, opened 2024-01-30
Make tests concurrency safe and faster

Murderlon:
Attempt at running package tests in parallel as well as parallel runners between Node.js versions.

**Update**: getting rate limiting issues when running Node.js runners in parallel.


Murderlon:
The real speed up would've been running the runners in parallel, but I'm getting rate limited on my GCS/S3 accounts then. Running the package tests in parallel shaved off one minute (5min->4min), which isn't shocking but it's something.

PR #583, opened 2024-02-15
fix: allow lockDrainTimeout to be configurable

fenos:
This PR simply allows customizing the `lockDrainTimeout` option.

When a request gets canceled by another lock, this option allow the cancelled request to wait until the `lockDrainTimeout` before starting the shutdown procedure, this is useful when we want to give a bit more time to a request to try and finish gracefully. 

On the cancelling side, the lock will be held for this same period

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -288,7 +288,7 @@ export class Server extends EventEmitter {
       abortWithDelayController.signal.removeEventListener('abort', onDelayedAbort)
       setTimeout(() => {
         requestAbortController.abort(err)
-      }, 3000)
+      }, this.options.lockDrainTimeout || 3000)
     }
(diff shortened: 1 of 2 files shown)


packages/server/src/server.ts
```diff
@@ -288,7 +288,7 @@ export class Server extends EventEmitter {
       abortWithDelayController.signal.removeEventListener('abort', onDelayedAbort)
       setTimeout(() => {
         requestAbortController.abort(err)
-      }, 3000)
+      }, this.options.lockDrainTimeout || 3000)
```
Murderlon, line 291: I would like to make it a habit to set default values in the `constructor`, then we'll have a single overview of all the default values.

Murderlon (commented):
LGTM. Just missing docs and a changeset (`npx changeset`).

Issue #598, opened 2024-04-11
How to get final name back in client?


tbunique:
Hello,

```
new Server({
    path: '/files',
    datastore: s3Store,
    namingFunction(req, metadata) {
        const newName = removeSuperfluousChars(metadata.filename);
        return newName;
    },
    onUploadCreate(req, res, upload) {
        const isAuthorised = checkSignature(upload.metadata);

        if (isAuthorised) {
            return res;
        }

        throw {
            status_code: 400
        };
    },
    onUploadFinish(req, res, upload) {
        deleteInfoFile(upload.metadata.filename);
        // res.setHeader('file-location', upload.id);
        return res;
    }
});
```

I'm using TUS to enable our clients to upload files into our S3 but there is something I'm missing. I use `namingFunction` to remove any weird characters the client would have in their name before storing it.
But I can't find how to send that clean name back to the client. I've tried modifying the headers in the `onUploadFinish method` and the header is correctly sent but CORS blocks the browser from reading the new name. I've checked the TUS source code and it seems I can't add anything to the EXPOSED_HEADERS.

I could move the renaming logic to the browser but I'd rather not the client being able to reintroduce weird characters.

Thanks in advance for any advice!

Murderlon:
Hi, you shouldn't use `namingFunction` to change the default unique ID to a non-unique file name. This will cause file corruption sooner or later if people upload files with the same name!

What probably makes most sense is to have #594, then you can change the file name in the `metadata` and the store will save that new metadata. You can keep the actual files uniquely generated. In tus in generally works like this, original file name in the `.info` or `.json` file and a unique file adjacent to it.

Performing a HEAD request will send back the metadata whenever the client likes to know it.

tbunique:
I see, thanks 👍 

I'm starting to think maybe tus is not the best for my use case. I want users to be able to overwrite existing files as they handle their own files. That's why I need to display the real file names without doing any HEAD requests (or querying my db where I would have stored the real hierarchy).

Murderlon:
You can override files, you just want to do it safely and explicitly. If you only use `namingFunction` without extra logic that puts each user's files in their own folder, a user can corrupt other user's files.

You can remove the old file in `onUploadFinish` to "override" it. But as of now, you would need to perform a HEAD to see the changes. I think it would be nice to send back `Upload-Metadata` in the POST in a future version of the tus spec.

tbunique:
I see what you mean.

I do have that extra logic elsewhere:

[...]

Issue #593, opened 2024-03-26
Is there a way to pass variables between the `onUploadCreate` and the `onUploadFinish` functions?


sammibajrami:
I have figured out a way to do it but it's very awkward and also involves changing and adding a few lines in the  `@tus/server/dist/handlers/PostHandler.js` file.

Basically the way that I have it working is that you update the "Upload-Metadata" header of the `res` object in the `onUploadCreate` function.
This header is then parsed in the `PostHandler.js` file and the result overwrites the `upload.metadata` variable, and then the upload object passes the extra metadata to the `onUploadFinish` function.

Any chance that there is a better way already built in that I'm missing?

Thank you in advance!

Murderlon:
Hi, this is an [XY problem](https://meta.stackexchange.com/questions/66377/what-is-the-xy-problem/66378#66378). What is the problem you are trying to solve? Are you only communicating between hooks or does the client also need to know this value?

Maybe you can just do it in server memory with a `Map`. Or if it needs to survive server restarts, put it in something like Redis?

sammibajrami:
I'm sorry, here is the problem:
I need to pass a variable from the `onUploadCreate` function to the `onUploadFinish` function. It does not need to be sent back to the client.

Is there a way to do this?

If not, then I'll look into your other suggestions, thank you!

Murderlon:
I don't know the requirements and why you are trying to pass that variable. There are many ways to write/read to different things depending on what you are doing, but those all fall into your hands and setup, there is no built-in way of doing this.

sammibajrami:
I'm sorry for being unclear.

I'm using the `onUploadCreate` function to validate the upload request, a process that generates a unique ID that I need to get in the `onUploadFinish` function.

Redis would be a way to do this, but I was hoping that I could simply update the metadata of the upload, but it sounds like that is not an option, correct?

Murderlon:
If you tell me why you need the ID to be accessible by the `onUploadFinish` hook, we can see if there are alternative approaches.

You can't change the metadata in `onUploadCreate` in a way that's persisted in the store too. We may want to support that at some point.

sammibajrami:
Once the upload is completed I create an entry for it, that includes the unique ID, in a separate database.

I'd be happy to share the code that I wrote to make the changes to the `upload` object persistent or even make a pull request if you think it'd help.
I've got it down to just adding two lines and changing two lines in the `tus-node-server/packages/server/src/handlers/PostHandler.ts` file.

[...]

PR #599, opened 2024-04-15
@tus/server: allow metadata changes in onUploadCreate

### LINKED ISSUE #594: Allow hooks to override metadata
Allow hooks to override metadata, only metadata not other `Upload` properties, in a backwards compatible way to precent breaking changes in the hook design. New metadata should probably be validated as well.

Murderlon:
Closes #594 
Ref: #598 
Ref: #593

The types are atrocious because TS only understands type narrowing with `instanceof` but we can't use that because tests use mocked versions of `http.ServerResponse`.

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -140,11 +140,12 @@ finished uploads. (`boolean`)
 #### `options.onUploadCreate`
 
 `onUploadCreate` will be invoked before a new upload is created.
-(`(req, res, upload) => Promise<res>`).
+(`(req, res, upload) => Promise<{ res: http.ServerResponse, metadata?: Record<string, string>}>`).
 
-If the function returns the (modified) response, the upload will be created. You can
-`throw` an Object and the HTTP request will be aborted with the provided `body` and
-`status_code` (or their fallbacks).
+- If the function returns the (modified) response the upload will be created.
+- You can optionally return `metadata` which will override (not merge!) `upload.metadata`.
+- You can `throw` an Object and the HTTP request will be aborted with the provided `body`
+  and `status_code` (or their fallbacks).
 
 This can be used to implement validation of upload metadata or add headers.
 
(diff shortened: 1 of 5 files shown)


packages/server/README.md
```diff
@@ -140,11 +140,12 @@ finished uploads. (`boolean`)
 #### `options.onUploadCreate`
 
 `onUploadCreate` will be invoked before a new upload is created.
-(`(req, res, upload) => Promise<res>`).
+(`(req, res, upload) => Promise<{ res: http.ServerResponse, metadata?: Record<string, string>}>`).
 
-If the function returns the (modified) response, the upload will be created. You can
-`throw` an Object and the HTTP request will be aborted with the provided `body` and
-`status_code` (or their fallbacks).
+- If the function returns the (modified) response the upload will be created.
```
Acconut, line 145: Does this mean that this change is backwards compatible? I.e. `onUploadCreate` implementations from before will continue working?

packages/server/README.md
```diff
@@ -445,13 +446,14 @@ const {Server} = require('@tus/server')
...
   async onUploadCreate(req, res, upload) {
-    const {ok, expected, received} = validateMetadata(upload)
+    const {ok, expected, received} = validateMetadata(upload) // your logic
     if (!ok) {
       const body = `Expected "${expected}" in "Upload-Metadata" but received "${received}"`
       throw {status_code: 500, body} // if undefined, falls back to 500 with "Internal server error".
     }
-    // We have to return the (modified) response.
-    return res
+    // You can optionally return metadata to override the upload metadata
```
Acconut, line 454:
```suggestion
    // You can optionally return metadata to override the upload metadata, such as `{ storagePath: "/upload/123abc..." }`
```

How about including an example in here?

Acconut (approved):
Looks great!

packages/server/README.md
Murderlon: Correct, hence the ugly temporary code/types. Both versions are also tested (another existing test not visible in the diff)

packages/server/README.md
Murderlon:
```suggestion
    // You can optionally return metadata to override the upload metadata,
    // such as `{ storagePath: "/upload/123abc..." }`
```

PR #605, opened 2024-04-30
@tus/server: introduce `POST_RECEIVE_V2`

### LINKED ISSUE #528: Call `POST-RECEIVE` while streaming instead of when PATCH is finished
Unless I'm missing something there doesn't seem to be a hook or event to track upload progress server side. [tusd](https://github.com/tus/tusd/blob/main/docs/hooks.md#list-of-available-hooks) has a hook called `post-receive` which allows you to log upload progress. Is there a way with `tus-node-server` to do something similar?

More info on use case. I'm using tus to upload the files. Once the upload is complete I will pass it off to another service to do the post processing and final storage. I'd like to have a task tracking wrapper to alert the user to the upload and post processing progress.

Murderlon:
Closes #528

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -247,6 +247,12 @@ server.on(EVENTS.POST_RECEIVE, (req, res, upload => {})
 Called every [`postReceiveInterval`](#optionspostreceiveinterval) milliseconds for every
 upload while it‘s being written to the store.
 
+This means you are not guaranteed to get (all) events for an upload. For instance if
+`postReceiveInterval` is set to 1000ms and an upload takes 500ms or when the upload takes
+2500ms, you would get the offset at 2000ms, not from 2500ms.
+
+Use `POST_FINISH` if you need to know when an upload is done.
+
 ```js
 const {EVENTS} = require('@tus/server')
 // ...
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -69,6 +69,11 @@ Max file size (in bytes) allowed when uploading (`number` |
 (`(req, id: string | null) => Promise<number> | number`)). When providing a function
 during the OPTIONS request the id will be `null`.
 
+#### `options.postReceiveInterval`
+
+Interval in milliseconds for sending progress of an upload over
+[`POST_RECEIVE_V2`](#eventspost_receive_v2) (`number`).
+
 #### `options.relativeLocation`
 
 Return a relative URL as the `Location` header to the client (`boolean`).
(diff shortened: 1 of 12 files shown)


packages/server/README.md
```diff
@@ -228,14 +233,32 @@ server.on(EVENTS.POST_CREATE, (req, res, upload => {})
 
 #### `POST_RECEIVE`
 
-Called every time a `PATCH` request is handled.
+Called every time an upload has been finished writing to the store. See also
+`POST_RECEIVE_V2`.
```
Acconut, line 237:
```suggestion
Called every time an upload has been finished writing to the store. This event is emitted
whenever the request handling is completed, whereas the `POST_RECEIVE_V2` event is emitted
while the request is being handled. 
```

This might explain the difference between the two events better. I am also thinking if we should choose a different name for this new event, like `DURING_RECEIVE`. It would indicate that it is emitted while the server is receiving data.

In my experience, a v2 indicator is often interpreted as a sign that this thing is new, better and should be preferred over the previous v1. Is this the case here as well? If not, we might want to drop the v2 suffix in favor of a more expressive event name.

I know that you are using these event names to be similar to tusd's hooks, but our decision to use `post-receive` there is also a bit questionable in hindsight. So we don't have to restrict ourselves to whatever odd name we came up with in tusd.

packages/server/README.md
```diff
@@ -228,14 +233,32 @@ server.on(EVENTS.POST_CREATE, (req, res, upload => {})
...
 ```
 
+#### `POST_RECEIVE_V2`
+
+Called every [`postReceiveInterval`](#optionspostreceiveinterval) milliseconds for every
+upload while it‘s being written to the store.
+
+This means you are not guaranteed to get (all) events for an upload. For instance if
+`postReceiveInterval` is set to 1000ms and an upload takes 500ms or when the upload takes
+2500ms, you would get the offset at 2000ms, not from 2500ms.
```
Acconut, line 252:
```suggestion
This means you are not guaranteed to get (all) events for an upload. For instance if
`postReceiveInterval` is set to 1000ms and an PATCH request takes 500ms, no event is emitted.
If the PATCH request takes 2500ms, you would get the offset at 2000ms, but not at 2500ms.
```

packages/server/src/handlers/BaseHandler.ts
```diff
@@ -149,6 +149,16 @@ export class BaseHandler extends EventEmitter {
         reject(err.name === 'AbortError' ? ERRORS.ABORTED : err)
       })
 
+      const postReceive = throttle((offset: number) => {
+        this.emit(EVENTS.POST_RECEIVE_V2, req, {...upload, offset})
+      }, this.options.postReceiveInterval)
```
Acconut, line 154: Should we specify the leading or trailing options here to get the behavior that is mentioned in the docs? https://lodash.com/docs/#throttle Or do the defaults already provide the behavior we need?

Acconut (approved):
Nice work! I only have a few minor comments, but otherwise, this look good to go!

packages/server/src/handlers/BaseHandler.ts
Murderlon: I think I prefer the defaults, which means the first event is immediately fired and if another event comes in within the wait period, it's fired after the timeout.

packages/server/README.md
Acconut: Ok, then let's leave it like this.

packages/server/src/handlers/BaseHandler.ts
Acconut: Ok, but is the documentation correct then? It currently says 

> This means you are not guaranteed to get (all) events for an upload. For instance if
`postReceiveInterval` is set to 1000ms and an PATCH request takes 500ms, no event is emitted.

But if we also trigger the event on a leading edge, we would get an event at 0m, wouldn't we?

Murderlon: Good catch. Perhaps the leading edge is not as valuable, which would be fired almost instantly (and we have `onIncomingRequest`) 🤔 If we set `leading` to `false` the docs are correct.

Acconut: > If we set `leading` to `false` the docs are correct.

Sounds good 👍

PR #612, opened 2024-05-05
@tus/gcs-store: fix multichunk uploads

### LINKED ISSUE #548: CRC errors using @google-cloud/storage 7.7.0
gcs-store package.json allows any version of @google-cloud/storage, but demo builds with 6.12.0 and works (once patched). 

Building with 7.7.0 of @google-cloud/storage gives error 
CRC32 missing from the final portion of a resumed upload

This was on resumable with a chunk size.

forcing a version 6 revision of storage fixes the issue for now as a work around.

netdown:
Fixes #548.

Multichunk uploads are broken starting from @google-cloud/storage v7.6. The reason is that [this PR](https://github.com/googleapis/nodejs-storage/pull/2333/files#diff-bc9705d0f7a567399044dfc66ccc82d4d9aa1cff116842a0094d54e463c9ecbc) introduces a CRC32 check if `File.createWriteStream` is called with an `offset` param set.

Since tus-node-server does not rely on the internal resumable upload feature of GCS (which was actually broken below 7.6), there is no need to create the stream with the offset param at all. See docs: https://googleapis.dev/nodejs/storage/latest/global.html#CreateWriteStreamOptions


Murderlon (approved):
Ideally a test would have caught this but I need to restructure the e2e tests so more tests can be reused.

Issue #606, opened 2024-04-30
onUploadFinish should be able to alter response body


netdown:
Hi,
If I'm not mistaken, currently there is no way to set a response body when the upload succeeds (if I throw with 204 status code, some headers wont be sent), the most I can do is set some custom headers. I think its common that you need to send a detailed json response after processing the file.

In tusd, the pre-finish hook is much more flexible as it expects an object from the hook, i.e.:
`{
    "HTTPResponse": {
        "Body": "{\"message\":\"done\"}",
    }
},`

I'm willing to make a PR, but first I'd like to discuss the ideal solution.

I think the onUploadFinish return type could be changed to `ServerResponse|{ res: ServerResponse, body?: string|ArrayBuffer, headers?: Record<string, string|number> }`. Currently, body is not passed to res.write at all, and header arrays could be easily merged. This would introduce a minimal modification in `packages/server/src/handlers/PatchHandler.ts`, around line 112.

Murderlon:
Hi, I'm surprised tusd allows it because a `204` PATCH is not allowed to have a body (cc @Acconut). And you are not allowed to change the status code:

> The Server MUST acknowledge successful PATCH requests with the 204 No Content
> — https://tus.io/protocols/resumable-upload#patch

Regarding headers, have you tried `res.setHeader()` in your hook?

If we want to change the return type it would have to be backwards compatible as done here: #599. But let's first discuss whether it makes sense at all.

Acconut:
> Hi, I'm surprised tusd allows it because a `204` PATCH is not allowed to have a body (cc @Acconut). And you are not allowed to change the status code:
> 
> > The Server MUST acknowledge successful PATCH requests with the 204 No Content
> > — https://tus.io/protocols/resumable-upload#patch

In hindsight, enforcing a 204 is a bit too restrictive and we should also allow 200 as most clients will accept both, in my experience. If people want to have custom response bodies for PATCH requests and their client implementations support that, I don't see a reason why tusd should interfere here.

netdown:
MDN web docs states that _The protocol allows user agents to vary in how they process such responses. This is observable in persistent connections, where the invalid body may include a distinct response to a subsequent request. Apple Safari rejects any such data. Google Chrome and Microsoft Edge discard up to four invalid bytes preceding a valid response. Firefox tolerates in excess of a kilobyte of invalid data preceding a valid response._

In this case there must be support for using 200. I think its fine that the status code must be changed manually.

PR #615, opened 2024-05-07
@tus/server: allow onUploadFinish hook to override response data

netdown:
Follow-up on #606. Changes based on #599.

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -159,7 +159,6 @@ This can be used to implement validation of upload metadata or add headers.
 `onUploadFinish` will be invoked after an upload is completed but before a response is
 returned to the client (`(req, res, upload) => Promise<{ res: http.ServerResponse, status_code?: number, headers?: Record<string, string | number>, body?: string }>`).
 
-- If the function returns the (modified) response, the upload will finish.
 - You can optionally return `status_code`, `headers` and `body` to modify the response.
   Note that the tus specification does not allow sending response body nor status code
   other than 204, but most clients support it. Use at your own risk.
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -157,11 +157,14 @@ This can be used to implement validation of upload metadata or add headers.
 #### `options.onUploadFinish`
 
 `onUploadFinish` will be invoked after an upload is completed but before a response is
-returned to the client (`(req, res, upload) => Promise<res>`).
+returned to the client (`(req, res, upload) => Promise<{ res: http.ServerResponse, status_code?: number, headers?: Record<string, string | number>, body?: string }>`).
 
-If the function returns the (modified) response, the upload will finish. You can `throw`
-an Object and the HTTP request will be aborted with the provided `body` and `status_code`
-(or their fallbacks).
+- If the function returns the (modified) response, the upload will finish.
+- You can optionally return `status_code`, `headers` and `body` to modify the response.
+  Note that the tus specification does not allow sending response body nor status code
+  other than 204, but most clients support it. Use at your own risk.
+- You can `throw` an Object and the HTTP request will be aborted with the provided `body`
+  and `status_code` (or their fallbacks).
 
 This can be used to implement post-processing validation.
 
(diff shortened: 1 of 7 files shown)


packages/server/README.md
```diff
@@ -157,11 +157,14 @@ This can be used to implement validation of upload metadata or add headers.
 #### `options.onUploadFinish`
 
 `onUploadFinish` will be invoked after an upload is completed but before a response is
-returned to the client (`(req, res, upload) => Promise<res>`).
+returned to the client (`(req, res, upload) => Promise<{ res: http.ServerResponse, status_code?: number, headers?: Record<string, string | number>, body?: string }>`).
 
-If the function returns the (modified) response, the upload will finish. You can `throw`
-an Object and the HTTP request will be aborted with the provided `body` and `status_code`
-(or their fallbacks).
+- If the function returns the (modified) response, the upload will finish.
```
Murderlon, line 162:
```suggestion
```

packages/server/src/handlers/PostHandler.ts
```diff
@@ -178,13 +197,18 @@ export class PostHandler extends BaseHandler {
...
           creation.getTime() + this.store.getExpiration()
         ).toUTCString()
       }
     }
 
-    const writtenRes = this.write(res, 201, {Location: url, ...headers})
+    const writtenRes = this.write(
+      res,
+      responseData.status,
+      Object.assign({Location: url}, responseData.headers),
```
Murderlon, line 209: I think it should be the other way around so hooks can't override the `Location.
```suggestion
      {...responseData.headers, Location: url},
```

packages/server/src/types.ts
```diff
@@ -107,7 +107,8 @@ export type ServerOptions = {
 
   /**
    * `onUploadFinish` will be invoked after an upload is completed but before a response is returned to the client.
-   * If the function returns the (modified) response, the upload will finish.
+   * If the function returns the (modified) response, the upload will finish. You can optionally return `status_code`, `headers` and `body` to modify the response.
```
Murderlon, line 110:
```suggestion
   * You can optionally return `status_code`, `headers` and `body` to modify the response.
```

Murderlon (commented):
Thanks for working on this!

packages/server/src/handlers/PostHandler.ts
netdown: As we are aiming at complete control over the response (including the status code), I would let the user override the location header. Possibly, in a customized implementation the resource url is not relevant at all, only a http 200 json response is required.

Murderlon: That doesn't make sense because we have `generateUrl` for that, which should be used. It should be locked in here.

netdown: I see. Initially I thought of being able to unset the location header, but I realized that its still not possible this way. Since location is only valid for 201 and 3xx, it would be good to support this. What do you think?

Murderlon: How about conditionally setting it based on the status code?

netdown: Sounds fine, I pushed the changes

packages/server/src/handlers/PostHandler.ts
```diff
@@ -203,10 +203,15 @@ export class PostHandler extends BaseHandler {
       }
     }
 
+    //Only append Location header if its valid for the final http status (201 or 3xx)
+    if (responseData.status === 201 || responseData.status.toString().at(0) === '3') {
```
Murderlon, line 207:
```suggestion
    if (responseData.status === 201 || responseData.status >= 300 && responseData.status < 400) {
```

Murderlon (commented):
Thanks for making a test as wel

andreineamtu:
Thank you for the test too! From there I understood that it's important to include the "res" object too so it works!
Also that line "If the function returns the (modified) response, the upload will finish." in the README file is very important, but in the latest version is not present anymore. Could you put it back?

Murderlon:
So it wasn't clear that you have to return `res` in the object? In the next major it will be removed but for now it's needed.

PR #616, opened 2024-05-07
@tus/server: add GCS locker

netdown:
This PR is not complete yet, it misses unit tests (the code is tested), readme updates and changeset. Despite all that, I would like to ask you to review my approach first so I won't write needless tests. I have documented the process in detail, but feel free to ask questions.

### DIFF
--- a/packages/server/src/lockers/GCSLocker.ts
+++ b/packages/server/src/lockers/GCSLocker.ts
@@ -1,0 +1,251 @@
+import {ERRORS, Lock, Locker, RequestRelease} from '@tus/utils'
+import {Bucket, File} from '@google-cloud/storage'
+import EventEmitter from 'node:events'
+
+/**
+ * Google Cloud Storage implementation of the Locker mechanism with support for distribution.
+ * For general information regarding Locker, see MemoryLocker.
+ *
+ * Locking is based on separate .lock files created in a GCS bucket (presumably the same as the upload destination, but not necessarily). Locker distribution is accomplished through metadata of the lockfile. After a lock file is created, we regularly check if its metadata was modified by another process (i.e. to request releasing the resource). To avoid resources being locked forever, each lock is created with an expiration time (also stored as metadata).
+ *
+ * Lock file health - possible states of a lock file:
+ * - non-existing (not locked)
+ * - active (locked)
+ * - requested to be released (locked, but should be released soon)
+ * - expired (not locked)
+ *
+ * Acquiring a lock:
+ * - If the lock file does not exist yet, create one with an expiration time and start watching it (see below)
+ * - If the lock file already exists
+ * -- If it has expired, treat it as non existing and overwrite it
+ * -- If it is active, request releasing the resource by updating the lockfile's metadata, then retry locking with an exponential backoff
+ *
+ * Releasing a lock:
+ * Stop the watcher and delete the lock file.
+ *
+ * Watching a lock (performed in every `watchInterval` ms):
+ * - If the lock file does not exist anymore, stop the watcher
+ * - If the lock file still exists, fetch its metadata
+ * -- If there is a release request in the metadata, call the cancel handler and stop the watcher
+ * -- If the lock has expired, call the cancel handler and stop the watcher
+ *
+ * (The process might be improved by introducing a secondary expiration time which gets updated by each watcher interval. This way we'll immediately know if the process which locked the resource has unexpectedly terminated and the resource should be released. Currently, only the `unlockTimeout` catches this scenario. However, this would introduce way more requests to GCS only for better handling of an extraordinary situation.)
+ *
+ */
+
+export interface GCSLockerOptions {
+  /**
+   * The bucket where the lock file will be created. No need to match the upload destination bucket.
+   */
+  bucket: Bucket
+  /**
+   * Maximum time (in milliseconds) to wait for an already existing lock to be released, else deny acquiring the lock.
+   */
+  acquireLockTimeout?: number
+  /**
+   * Maximum lifetime (in milliseconds) of a lock. Processes may unexpectedly quit, we need to make sure resources won't stay locked forever. Make sure this is a safe maximum, else the lock may be released while the resource is still being used.
+   */
+  unlockTimeout?: number
+  /**
+   * The amount of time (in milliseconds) to wait between lock file health checks. Larger interval results less requests to GCS, but generally more time to release a locked resource. Must be less than `acquireLockTimeout`.
+   */
+  watchInterval?: number
+}
+
+export class GCSLocker implements Locker {
+  events: EventEmitter
+  bucket: Bucket
+  lockTimeout: number
+  unlockTimeout: number
+  watchInterval: number
+
+  constructor(options: GCSLockerOptions) {
+    this.events = new EventEmitter()
+    this.bucket = options.bucket
+    this.lockTimeout = options.acquireLockTimeout ?? 1000 * 30
+    this.unlockTimeout = options.unlockTimeout ?? 1000 * 600
+    this.watchInterval = options.watchInterval ?? 1000 * 10
+
+    if (this.watchInterval < this.lockTimeout) {
+      throw new Error('watchInterval must be less than acquireLockTimeout')
+    }
+  }
+
+  newLock(id: string) {
+    return new GCSLock(id, this)
+  }
+}
+
+class GCSLock implements Lock {
+  constructor(private id: string, private locker: GCSLocker) {}
+
+  async lock(requestRelease: RequestRelease): Promise<void> {
+    const abortController = new AbortController()
+
+    const lock = await Promise.race([
+      this.waitForLockTimeoutOrAbort(abortController.signal),
+      this.acquireLock(requestRelease, abortController.signal),
+    ])
+
+    abortController.abort()
+
+    if (!lock) {
+      throw ERRORS.ERR_LOCK_TIMEOUT
+    }
+  }
+
+  async unlock(): Promise<void> {
+    const lockFile = new GCSLockFile(this.locker, this.id)
+
+    if (!(await lockFile.isLocked())) {
+      throw new Error('Releasing an unlocked lock!')
+    }
+
+    await lockFile.delete()
+  }
+
+  protected async acquireLock(
+    cancelHandler: RequestRelease,
+    signal: AbortSignal,
+    attempt = 0
+  ): Promise<boolean> {
+    if (signal.aborted) {
+      return false
+    }
+
+    const lockFile = new GCSLockFile(this.locker, this.id)
+
+    if (!(await lockFile.isLocked())) {
+      //The id is not locked yet - create a new lock file on GCS, start watching it
+      await lockFile.write(cancelHandler)
+
+      return true
+    } else {
+      //The id is already locked, we need to request releasing the resource
+      await lockFile.requestRelease()
+
+      //Try to acquire the lock again
+      return await new Promise((resolve, reject) => {
+        //On the first attempt, retry after current I/O operations are done, else use an exponential backoff
+        const waitFn = (then: () => void) =>
+          attempt > 0
+            ? setTimeout(then, (attempt * this.locker.watchInterval) / 3)
+            : setImmediate(then)
+
+        waitFn(() => {
+          this.acquireLock(cancelHandler, signal, attempt + 1)
+            .then(resolve)
+            .catch(reject)
+        })
+      })
+    }
+  }
+
+  protected waitForLockTimeoutOrAbort(signal: AbortSignal) {
+    return new Promise<boolean>((resolve) => {
+      const timeout = setTimeout(() => {
+        resolve(false)
+      }, this.locker.lockTimeout)
+
+      const abortListener = () => {
+        clearTimeout(timeout)
+        signal.removeEventListener('abort', abortListener)
+        resolve(false)
+      }
+      signal.addEventListener('abort', abortListener)
+    })
+  }
+}
+
+class GCSLockFile {
+  protected fileId: string
+  protected lockFile: File
+  protected unlockTimeout: number
+  protected watchInterval: number
+  protected watcher: NodeJS.Timeout | undefined
+
+  constructor(locker: GCSLocker, fileId: string) {
+    this.fileId = fileId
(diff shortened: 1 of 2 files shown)


Murderlon (commented):
Thanks for putting in the time to contribute!

I'm not an expert on (distributed) locking, but conceptually I think GCS storage as a locker only makes sense if you're already deploying your server within GCS infrastructure (so it's faster) and you have a bucket in the region where the uploads happen. My assumption is if those conditions aren't met, things will be slow? AFAIK GCS has strong consistency within the same region but eventual consistency for multi-region.

Maybe you can elaborate on your use case?

netdown:
Indeed I haven't even thought about using this locker with a store other than GCS. In my case, the storage bucket and the locker bucket is the same, and I think the only case they should be separated is when the storage bucket is not in standard storage class. 

Anyways, I'm not sure i.e. Firestore would greatly overperform GCS in case of different storage. Regarding region latency, the user should be aware of that and choose a suitable region. Of course a redis based implementation would be much better, but this may be a considerable alternative until thats not implemented.

Shall I move this locker to the gcs-store package to suggest the primary application?

Acconut (commented):
This is interesting because such approaches would allow tus server to implement lockers directly on top of cloud storages instead of using external tools like Redis. However, I would like to see some evidence that this approach actually provides exclusive access to uploads. Is there some blog post that looked into the mechanisms at play here? Are all involved operations [strongly consistent](https://cloud.google.com/storage/docs/consistency)?

netdown:
GCS is strongly consistent, but indeed concurrency was not ensured in my previous approach. I have reworked the code based on [this article](https://www.joyfulbikeshedding.com/blog/2021-05-19-robust-distributed-locking-algorithm-based-on-google-cloud-storage.html).

Note that I had to upgrade @google-cloud/storage because previous version was missing a type export. Also, this feature should be moved to a separate package or into gcs-store, as I'm importing from @google-cloud/storage.

Murderlon:
Really nice article, thanks for sharing. It does also say this:

> A distributed lock like this is best suited for batch workloads. Such workloads typically take seconds to tens or even hundreds of seconds.

But here we are using it for individual uploads, not batches. Or even smaller with a resumed uploads (or where a client sets `chunkSize` on the client). It's probably fine, but hard to tell when this becomes a bottle neck. Have you already tried this locker in production? Ideally we know if this behaves well before merging.

netdown:
For the last 10 days it has been running in production without problems. We have about 5000 uploads per day. In e2e tests it was indeed slightly slower for 140 files compared to xhr, but I could easily compensate this by increasing the number of parallel uploads. If I measure individual uploads, the time elapsed between lock and unlock is mostly 20-400 ms in case of memory locker, and 300-400 for gcs locker.

Murderlon:
That's great to hear! I'm in favor adding this into the package then.

packages/server/src/lockers/GCSLockFile.ts
```diff
@@ -0,0 +1,129 @@
...
+      metadata,
+    })
+    this.currentMetaGeneration = 0
+  }
+
+  /**
+   * Fetch metadata of the lock file.
+   */
+  public async getMeta() {
+    return (await this.lockFile.getMetadata())[0] as GCSLockFileMetadata
```
Murderlon, line 58: Are we sure this can never return `undefined` or something else?

packages/server/src/lockers/GCSLockFile.ts
```diff
@@ -0,0 +1,129 @@
...
+    return releaseRequestExists
+  }
+
+  /**
+   * Delete our own lockfile if it still exists.
+   */
+  public async deleteOwn() {
+    try {
+      await this.deleteReleaseRequest()
+      await this.lockFile.delete({ifGenerationMatch: this.currentMetaGeneration})
```
Murderlon, line 92: Would it make sense to do this in parallel?

Murderlon (commented):
Overall this looks very good! Also happy with the extensive code comments.

Some things needed:

- The build is currently failing
- We need to update the `peerDependencies` to not allow any version.
- Docs. We should also talk about when to (not) use this lock and the things to watch out for, such as what values to set for the ttl and watch interval. 
- A test similar to this:
https://github.com/tus/tus-node-server/blob/a0f9da1450700d97a2e77d791e175ce3adaa1b78/test/e2e.test.ts#L1045-L1145

If you need help with any of these let me know.

Acconut:
> GCS is strongly consistent, but indeed concurrency was not ensured in my previous approach. I have reworked the code based on [this article](https://www.joyfulbikeshedding.com/blog/2021-05-19-robust-distributed-locking-algorithm-based-on-google-cloud-storage.html).

Thank you for the article, I will have a look at it! I am wondering if S3 has similar capabilities and a locker can be implemented nowadays ontop of it as well.

netdown:
Yes, but I've been busy the last few weeks and I expect the same at least until July. Feel free to complete the PR if you have the time.

packages/server/src/lockers/GCSLocker.ts
```diff
@@ -0,0 +1,155 @@
...
+            .catch(reject)
+        })
+      })
+    }
+
+    return true
+  }
+
+  protected waitForLockTimeoutOrAbort(signal: AbortSignal) {
+    return new Promise<boolean>((resolve) => {
```
Acconut, line 142: Node.js has a built-in async version of setTimeout: https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options

packages/server/src/lockers/GCSLockFile.ts
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Request releasing the lock from another instance. As metadata edits are only prohibited for the owner (so it can keep track of metageneration), we write to a separate file.
+   */
+  public async requestRelease() {
+    try {
+      await this.releaseFile.save('', {
+        preconditionOpts: {ifGenerationMatch: 0},
+      })
+    } catch (err) {
+      //Release file already created, no need to report
```
Acconut, line 107: I prefer check explicitly which error conditions should be ignored and then re-throwing all other errors. This applies to multiple try-catch block in this PR.

packages/server/src/lockers/GCSLockFile.ts
Acconut: It makes sense to delete the release request file first. Otherwise, the lock might be acquired by a new owner while the old release request file still exists.

packages/server/src/lockers/GCSLock.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    return true
+  }
+
+  /**
+   * Start watching the lock file - keep it healthy and handle release requests.
+   */
+  protected startWatcher(cancelHandler: RequestRelease) {
+    this.watcher = setInterval(() => {
+      const handleError = () => {
```
Acconut, line 94: The error handling here needs some extension. The error should probably be logged somewhere. In addition, if we cannot extend the lock's lifetime, we should stop all processing of the resource once the locker expires. This is probably hard to implement with the current lock interface but this is also something I would like to address with tusd in the future.

packages/server/src/lockers/GCSLock.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        return false
+      }
+    }
+  }
+
+  /**
+   * Release the lock - clear watcher and delete the file.
+   */
+  public async release() {
```
Acconut, line 59: Do we need some internal mutex to ensure that deleting a lock and extending it cannot run simultaneously and thus interfere with each other? Async requests can run concurrently.

Acconut (commented):
Apologies for my delayed review! I just read the accompanying [blog post](https://www.joyfulbikeshedding.com/blog/2021-05-19-robust-distributed-locking-algorithm-based-on-google-cloud-storage.html) and wanted to leave some comments about it first. Some additional background information can be found at [gcslock](https://github.com/mco-gh/gcslock), which was a previous GCS-based lock. The only valuable comment online I was able to find about the proposed algorithm is on [Lobsters](https://lobste.rs/s/tymd6a/robust_distributed_locking_algorithm) by [Aphyr](https://aphyr.com/about), who is quite experienced in testing distributed systems and databases. However, his comment was more about a general issue with distributed locks and not about this GCS-based approach in particular. The same [critique](https://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html) can also be applied to Redis-based locks and there is not much we can do on our end as far as I know.

The proposed algorithm on its own seems sound to me (although I am no expert). It relies on the storage offering strong consistency which is the case with GCS. While there are many S3-compatible storage I am not aware of any GCS-compatible storages. So we don't have to worry much about storages with a GCS-like interface that are not strongly consistent.

In addition, the propsed algorithm also provides "instant recovery from stale locks" if the lock was left stale by the same actor that now tries to acquire it. This functionality attaches an identity to each lock, which is dangerous for tus-node-server as we do not want two requests that are processed by the same tus-node-server instance to interfere with the same lock. This PR does not implement this feature but this difference to the blog post should still be noted in the code somewhere.
The author also acknowledges that this algorithm does not offer low-latency:

> A locking operation's average speed is in the order of hundreds of milliseconds.

This is probably fine for large file uploads, which are I/O-bound, but still work documenting somewhere.

Finally, while reading the article, I hoped that a similar approach might be possible for S3, but this does not seem possible at the first glance as it does not offer conditional writes like GCS does.

packages/server/src/lockers/GCSLock.ts
Murderlon: AFAIK GCS should already take care of concurrent access for us.

packages/server/src/lockers/GCSLockFile.ts
Murderlon: I'm not sure which error conditions this would be then

Acconut: We can only safely ignore the error if it's a `412 Precondition Failed` because then the release file exists already. All other errors (e.g. failure to reach GCS API entirely) should be thrown.

Murderlon:
**Update:**

- Moved GCSLocker into @tus/gcs-store
- Upgraded `@google-cloud/storage` and corrected the peer dependency
- Added tests
- Correctly set the expiry on file metadata.
- Fix setting `this.currentMetaGeneration` in `GCSLockFile.create()`. Without this, `requestRelease` was never called because the [meta generation didn't match](https://github.com/tus/tus-node-server/blob/823076b69db890dfc30435491a188751e3330d9b/packages/gcs-store/src/locker/GCSLockFile.ts#L78-L80).

```diff
- this.currentMetaGeneration = 0
+ this.currentMetaGeneration = (await this.getMeta()).metageneration
```

- Fixed double retrying.

Note that in `acquireLock` there is recursive retrying:

https://github.com/tus/tus-node-server/blob/b5e0bfb5167e69c862cd9dae6a8a242b51d5b471/packages/gcs-store/src/locker/GCSLocker.ts#L125-L134

but inside the `take` function, which is called inside `acquireLock`, we do recursion again:

https://github.com/tus/tus-node-server/blob/b5e0bfb5167e69c862cd9dae6a8a242b51d5b471/packages/gcs-store/src/locker/GCSLock.ts#L40-L52

From my understanding this wasn't needed and just caused for repetitive calls.

packages/gcs-store/src/locker/GCSLock.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    //Delete the lock file
+    await this.file.deleteOwn()
+  }
+
+  /**
+   * Check if the lock is healthy, delete if not.
+   * Returns TRUE if the lock is healthy.
+   */
+  protected async insureHealth() {
```
Acconut, line 76:
```suggestion
  protected async ensureHealth() {
```

Could this be a naming mistake? "ensure" seems more appropriate than "insure".

packages/gcs-store/src/locker/GCSLockFile.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.lockFile = bucket.file(name)
+    this.releaseFile = bucket.file(`${name}.release`)
+  }
+  /**
+   * Create the lockfile with the specified exp time. Throws if the file already exists
+   */
+  public async create(exp: number) {
+    const metadata = {
+      metadata: {exp},
+      // TODO: this does nothing?
```
Acconut, line 41: What about this TODO comment?

packages/gcs-store/src/locker/GCSLock.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      return true
+    } catch (err) {
+      log('failed creating lock file', err.code, err.message)
+      //Probably lock is already taken
+      const isHealthy = await this.insureHealth()
+
+      if (!isHealthy) {
+        log('lock not healthy. calling GCSLock.take() again')
+        return await this.take(cancelHandler)
```
Acconut, line 52: What happens when the lock is unhealthy and cannot be taken again? Is access to the upload resources on GCP then taken away? Since the locker cannot ensure exclusive access, saving uploaded data to GCS should be stopped.

packages/gcs-store/src/locker/GCSLockFile.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Delete our own lockfile if it still exists.
+   */
+  public async deleteOwn() {
+    try {
+      await this.deleteReleaseRequest()
+      await this.lockFile.delete({ifGenerationMatch: this.currentMetaGeneration})
+    } catch (err) {
+      //Probably already deleted, no need to report
```
Acconut, line 94: Only errors about the object not existing should be ignored. All other errors should be thrown.

packages/gcs-store/src/locker/GCSLockFile.ts
```diff
@@ -0,0 +1,136 @@
...
+    await this.lockFile.delete({ifMetagenerationMatch: metaGeneration})
+  }
+
+  /**
+   * Delete release request file of the lock if exists.
+   */
+  protected async deleteReleaseRequest() {
+    try {
+      await this.releaseFile.delete()
+    } catch (err) {}
```
Acconut, line 134: Only errors about the object not existing should be ignored. All other errors should be thrown.

Acconut (changes requested):
Only a few minor changes, otherwise LGTM 👍

PR #624, opened 2024-06-06
Add basic storage info to Upload model

### LINKED ISSUE #621: Upload using tus but serve files directly from s3 or using nginx without tus server
I am using tus along with uppy client in react app for uploading files which is working great.

**Objective:**
But I do not want to use tus at all for serving my uploaded files. I want to serve files directly from s3 or from local file system using nginx (whichever datastore I use) without having to pass the request through tus server. I can obviously put both s3 and local file system behind cdn of my choice. I feel the real benefit of tus is while uploading files and not after that. It would be great if there is an easier way to decouple the serving of files part.

**Issues:**

1. URL generated by tus is very different from the file name as well as the path where it is actually stored. So without tus server I can't map the url directly to actual file path. Tus basically generates a hash for the complete path of the file and then use that as the file id.
2. When I access the files in s3 directly then they are served using the contentType: `application/offset+octet-stream` and hence the file is downloaded instead of being rendered as image which I think I can override using the contentType metadata in tus client. But is there a way to do this in tus node server automatically for all files based on original file mimetype or something?
3. Is there any way I can force all the files to be stored using a custom name along with unique id and extension. Example: `some-file-name-${unique_id}.jpg`. The combination of both file name and unique id can help prevent conflicts between multiple files. Also, the extension can help in identifying the type of file quickly.
4. Since I want to serve files directly from respective datastore so can i somehow return a custom url (url of cdn or s3) once the file is uploaded.
5. Can I simply disable the generation of metadata files along with the actual file because I want to serve directly from s3 without the head request?

**Current Code:**

```
const datastore = new S3Store({
    partSize: 8 * 1024 * 1024, // Each uploaded part will have ~8MiB,
    s3ClientConfig: {
        bucket: process.env.AWS_BUCKET!,
        region: process.env.AWS_REGION!,
        credentials: {
            accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
            secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
        },
    },
});

const tusServer = new TusServer({
    path: options.path,
    respectForwardedHeaders: true,
    datastore: datastore,
    namingFunction(req) {
        const id = crypto.randomUUID();
        const userId = "userid"; // Parse user id from request
        return `users/${userId}/${id}`
    },
    generateUrl(req, { proto, host, path, id }) {
        return `${proto}://${host}${path}/${id}`
    },
    getFileIdFromRequest(req) {
        const match = reExtractFileID.exec(req.url as string)

        if (!match || options.path.includes(match[1])) {
            return;
        }

        return Buffer.from(match[1], 'base64url').toString('utf-8')
    },
});

server.addContentTypeParser('application/offset+octet-stream', (req, payload, done) => done(null));
server.all(`${options.path}`, (req, res) => { tusServer.handle(req.raw, res.raw); });
server.all(`${options.path}/*`, (req, res) => { tusServer.handle(req.raw, res.raw); });
```

Note that I am storing files of each user separately in their own directory.

Would really appreciate any help in this direction. Thanks!

Murderlon:
Closes #621 

Anything else we need to store in here?

cc @prolific @Acconut

### DIFF
--- a/.changeset/hungry-games-compete.md
+++ b/.changeset/hungry-games-compete.md
@@ -1,0 +1,9 @@
+---
+'@tus/file-store': minor
+'@tus/gcs-store': minor
+'@tus/s3-store': minor
+'@tus/utils': minor
+---
+
+Add basic storage information to the Upload model. You can now access `upload.storage`
+which has `type` (`file`, `s3`, `gcs`), `path`, and when applicable `bucket`.
--- a/packages/file-store/index.ts
+++ b/packages/file-store/index.ts
@@ -164,6 +164,7 @@ export class FileStore extends DataStore {
             offset: stats.size,
             metadata: file.metadata,
             creation_date: file.creation_date,
+            storage: {type: 'file', path: file_path},
           })
         )
       })
--- a/packages/file-store/index.ts
+++ b/packages/file-store/index.ts
@@ -60,10 +60,13 @@ export class FileStore extends DataStore {
    */
   async create(file: Upload): Promise<Upload> {
     const dirs = file.id.split('/').slice(0, -1)
+    const filePath = path.join(this.directory, file.id)
 
     await fsProm.mkdir(path.join(this.directory, ...dirs), {recursive: true})
-    await fsProm.writeFile(path.join(this.directory, file.id), '')
+    await fsProm.writeFile(filePath, '')
     await this.configstore.set(file.id, file)
+
+    file.storage = {type: 'file', path: filePath}
 
     return file
   }
--- a/packages/gcs-store/index.ts
+++ b/packages/gcs-store/index.ts
@@ -33,6 +33,8 @@ export class GCSStore extends DataStore {
 
       const gcs_file = this.bucket.file(file.id)
 
+      file.storage = {type: 'gcs', path: file.id, bucket: this.bucket.name}
+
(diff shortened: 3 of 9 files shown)


.changeset/hungry-games-compete.md
```diff
@@ -0,0 +1,9 @@
+---
+'@tus/file-store': minor
+'@tus/gcs-store': minor
+'@tus/s3-store': minor
+'@tus/utils': minor
+---
+
+Add basic storage information to the Upload model. You can now access `upload.storage`
+which has `type` (`file`, `s3`, `gcs`), `path`, and when applicable `bucket`.
```
Acconut, line 9: In tusd I have decided to use `key` for s3/gcs as that is the terminology that the providers also use. Maybe you want to consider this as well.

Murderlon: I saw that, but I don't want conditional types and conceptually we can treat them the same, you use it to know where an upload is stored.

Acconut: Ok, sounds good.

prolific:
@Murderlon  Looks good to me.

prolific (approved):
@Murderlon Looks good to me.

PR #636, opened 2024-07-25
@tus/server: add allowedCredentials and allowedOrigins options

idanto:
Hi,

Uppy client supports `withCredentials` parameter to allow sending tokens in cookie
https://uppy.io/docs/tus/#withcredentials

Tus Server doesn't allow any way to configure it; it seems like a missing configuration, and this MR is a suggestion on how to implement it.


Acconut:
Should the user also be able to enable this only for requests from certain origins? Allowing it from arbitrary origins is potentially dangerous.

idanto:
> Should the user also be able to enable this only for requests from certain origins? Allowing it from arbitrary origins is potentially dangerous.

@Acconut, isn't that what Access-Control-Allow-Origin should do in the CORS policy?

TBH, I don't see a way to limit it in Tus code. I see the opposite which looks like the client can override it by sending an origin header.

Acconut:
If tus-node-server does not yet have a way to restrict CORS access to certain origins, then we should probably add that as well.

idanto:
@Murderlon I agree in general, but this is a special case. Those are CORS options, which is a well-defined standard. In most of the server implementations, I know you usually have specific options to configure your CORS to prevent people from making mistakes (like typos and so on).

Murderlon:
My experience is that sooner or later people run into the limitations of these locked down options. For example, people with preview deployments, where the subdomain is dynamic, will not be able to use `string[]` for their origins.

If we make it `RegExp | string[]` I'm okay with keeping the options as proposed 👍

Murderlon:
@idanto are you still interested in getting this over the finish line?

idanto:
@Murderlon looks like after you merge master everything is green

Murderlon:
Only this needs to be resolved: https://github.com/tus/tus-node-server/pull/636#issuecomment-2257764146.

I can take a look later this week if you don't have time before then.

Murderlon:
Thanks for the contribution

idanto:
thanks for the help pushing it to the finish line @Murderlon

PR #641, opened 2024-08-01
Refactor TypeScript setup

remcohaszing:
This replaces turbo with TypeScript project references. Instead of `turbo run build`, now we can use `tsc --build`.

All projects now follow a consistent layout. The `src` directory is for source files. The `dist` directory is for compiled output. The `test` directory is for tests.

Tests still use `ts-node` and `mocha`. Every workspace now contains 2 TypeScript configuration files. `tsconfig.build.json` checks and compiles the source files. Every dependency must be referenced via the `references` property. The `tsconfig.json` file checks all files that are **not** part of the source, at the moment this is test files. Note that test files were previously not type checked at all.

The `test` workspace has a dependency on all packages’ `tsconfig.build.json` files. All workspace `tsconfig.json` files have a dependency on the `test` workspace. It’s worth noting that cyclic project references are not allowed, and we’re lucky tests are currently not part of the build.

The project root’s `tsconfig.json` served as a base configuration. It now serves as the solution file that references all workspaces. Every workspace can now be checked with `tsc --build` individually. The entire workspace with `tsc --build`. Instead, a new file, `tsconfig.base.json`, now serves as a base.

Some compiler options were changes:
- `module` is now set to `node16`.
- `esModuleInterop` was removed, as it is redundant with `module` `node16`.
- `skipLibCheck` was removed, as it’s good to know if there are any conflicts with third party types.
- `forceConsistentCasingInFileNames` was removed, as this is true by default.
- `moduleResolution` was removed, as this is configured automatically with `module` node16`.
- `resolveJsonModule` was removed, as it appears to be unused.
- `composite` was enabled, as it has to be for project references.

The `types` field was removed from `package.json` files. This is redundant if the declaration file names match the JavaScript file names.

Some additional things to consider after this PR:
- Enable `verbatimModuleSyntax` for stricter CJS interoperability checks.
- Enable `declarationMap` for _Go to Definition_ support.
- Enable `sourceMap` for source map support.
- Publish the source code to make declaration maps and source maps work for consumers.


Murderlon (approved):
Thanks for the PR!

PR #645, opened 2024-08-14
Basic support for Azure storage account upload

battajegroup:
I have added the basic support for Azure storage account upload. This is the basic implementation as the node sdk for Azure storage has some limitation for stream upload. Please have a look and let me know for any comments and suggestions. Thank you.

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -5,12 +5,14 @@
   "scripts": {
     "start": "node server.js",
     "start:gcs": "cross-env DATA_STORE=GCSDataStore node server.js",
-    "start:s3": "cross-env DATA_STORE=S3Store node server.js"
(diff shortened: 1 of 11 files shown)


Murderlon:
Sorry for the delay. I will take a look at it this week.

Murderlon (approved):
Thanks for the contribution! I made a handful of commits with tiny tweaks. The tests fail in CI because this PR gets the repository secrets at the time of opening the PR and I added them later. But I can confirm locally that all tests pass. 

Initial version will be `0.1.0` instead of `1.0.0` when published. When it has proven itself stable we can do a major.

battajegroup:
You're welcome. Thank you for the review and accepting the request.

PR #647, opened 2024-08-20
Replace Prettier and Eslint with Biome

Murderlon:
- Remove `packages/eslint-config-custom` and all its dependencies
- Remove eslintrc and eslint deps
- Remove prettierrc and prettier deps
- Remove `lint` and `format` commands from all packages. This is now only ran from root.


Murderlon:
@SocketSecurity ignore npm/@biomejs/biome@1.8.3

PR #655, opened 2024-09-16
@tus/server: add Content-Type and Content-Disposition headers on GetHandler.send response

jesusgoku:
The [tusd](https://github.com/tus/tusd) is the official implementation development in Go lang, and this adds `Content-Type` and `Content-Disposition` headers on response of `GetHandler`, this allows browser for:

- Recognize mime-type of file and render file inline or force download
- Suggest filename in save dialog

## References

- [GetFile in tusd source code](https://github.com/tus/tusd/blob/0339037f874ccea5c6abdbfb424541f39f2db565/pkg/handler/unrouted_handler.go#L984)

### DIFF
--- a/packages/server/src/handlers/GetHandler.ts
+++ b/packages/server/src/handlers/GetHandler.ts
@@ -11,6 +11,13 @@ export class GetHandler extends BaseHandler {
 
   reMimeType = /^[a-z]+\/[a-z0-9\-\+\.]+$/
 
+  /**
+   * mimeInlineBrowserWhitelist is a set containing MIME types which should be
+   * allowed to be rendered by browser inline, instead of being forced to be
+   * downloaded. For example, HTML or SVG files are not allowed, since they may
+   * contain malicious JavaScript. In a similar fashion PDF is not on this list
+   * as their parsers commonly contain vulnerabilities which can be exploited.
+   */
   mimeInlineBrowserWhitelist = new Set([
     'text/plain',
 
--- a/packages/server/src/handlers/GetHandler.ts
+++ b/packages/server/src/handlers/GetHandler.ts
@@ -9,7 +9,18 @@ import type {RouteHandler} from '../types'
 export class GetHandler extends BaseHandler {
   paths: Map<string, RouteHandler> = new Map()
 
-  reMimeType = /^[a-z]+\/[a-z0-9\-\+\.]+$/
+  /**
+   * reMimeType is a RegExp for check mime-type form compliance with RFC1341
+   * for support mime-type and extra parameters, for example:
+   *
+   * ```
+   * text/plain; charset=utf-8
+   * ```
+   *
+   * See: https://datatracker.ietf.org/doc/html/rfc1341 (Page 6)
+   */
+  reMimeType =
+    /^(?:application|audio|example|font|haptics|image|message|model|multipart|text|video|x-(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+))\/([0-9A-Za-z!#$%&'*+.^_`|~-]+)((?:[ 	]*;[ 	]*[0-9A-Za-z!#$%&'*+.^_`|~-]+=(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+|"(?:[^"\\]|\.)*"))*)$/
 
   /**
    * mimeInlineBrowserWhitelist is a set containing MIME types which should be
--- a/packages/server/src/handlers/GetHandler.ts
+++ b/packages/server/src/handlers/GetHandler.ts
@@ -34,6 +34,7 @@ export class GetHandler extends BaseHandler {
     'audio/webm',
     'audio/ogg',
 
+    'video/mp4',
     'video/webm',
     'video/ogg',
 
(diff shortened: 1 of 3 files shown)


Murderlon (commented):
Thanks for the PR and the well written tests! Only one Q

packages/server/src/handlers/GetHandler.ts
```diff
@@ -1,14 +1,38 @@
...
 
 import type http from 'node:http'
 import type {RouteHandler} from '../types'
 
 export class GetHandler extends BaseHandler {
   paths: Map<string, RouteHandler> = new Map()
 
+  reMimeType = /^[a-z]+\/[a-z0-9\-\+\.]+$/
+
+  mimeInlineBrowserWhitelist = new Set([
```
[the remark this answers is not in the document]
jesusgoku: ¡Hi @Murderlon! That's right, it actually makes no difference what type of files are uploaded, and therefore later downloaded. But as I mentioned, PR is based on the official [tusd](https://github.com/tus/tusd) implementation, and I leave below the [observation](https://github.com/tus/tusd/blob/0339037f874ccea5c6abdbfb424541f39f2db565/pkg/handler/unrouted_handler.go#L1046-L1052C18) they make regarding why they included only a couple of types in the whitelist:

> mimeInlineBrowserWhitelist is a map containing MIME types which should be
> allowed to be rendered by browser inline, instead of being forced to be
> downloaded. For example, HTML or SVG files are not allowed, since they may
> contain malicious JavaScript. In a similiar fashion PDF is not on this list
> as their parsers commonly contain vulnerabilities which can be exploited.
> The values of this map does not convey any meaning and are therefore just
> empty structs. 

It was my mistake not to have kept the same comment in the code, precisely to answer the question that reasonably arose. If you like, I can add this comment as part of the code?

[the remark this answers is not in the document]
Murderlon: I think it would be nice to add the comment yes! And perhaps add `video/mp4`?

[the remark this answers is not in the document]
jesusgoku: We're done! Add the comment and also add the mime-type for `video/mp4`. I've been [looking](https://github.com/search?q=repo%3Atus%2Ftusd%205e22d82ca5ea5c0cdb1a47ae02bbb4f0bc3c1250&type=code) for any discussion in a PR when those changes were introduced in tusd to see if they hadn't included it for some reason, but I couldn't find anything.

I've also supplemented the regular expression that validates the mime type form to be in accordance with [RFC1341](https://datatracker.ietf.org/doc/html/rfc1341) (Page 6) and allow the inclusion of parameters, useful for example to specify the content charset.

```
text/plain; charset=utf-8
```

packages/server/src/handlers/GetHandler.ts
```diff
@@ -1,14 +1,57 @@
...
+   * for support mime-type and extra parameters, for example:
+   *
+   * ```
+   * text/plain; charset=utf-8
+   * ```
+   *
+   * See: https://datatracker.ietf.org/doc/html/rfc1341 (Page 6)
+   */
+  reMimeType =
+    /^(?:application|audio|example|font|haptics|image|message|model|multipart|text|video|x-(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+))\/([0-9A-Za-z!#$%&'*+.^_`|~-]+)((?:[ 	]*;[ 	]*[0-9A-Za-z!#$%&'*+.^_`|~-]+=(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+|"(?:[^"\\]|\.)*"))*)$/
```
Murderlon, line 23: Is it worth it to support parameters as well? We now have a very complex, slow regex with lots of backtracking. If it doesn't add much value I think I prefer to stay with the simple version? What do you think?

jesusgoku: I understand the caution, this last regex is definitely much more intimidating than the previous one, but it is not significantly slower, less than 2ms on average, and this last one analyzing the list of 2138 mime types registered by the IANA.

And answering your question, I think it is important, especially for text files, to preserve and transmit the information of the character set used, and this can be rendered properly when displayed inline.

If you like, we can change to an intermediate version that does not try to match on the parameters, and only looks at the mime type part, after that "anything can come".

## Benchmarks

### Simple Regex
Simple Regex

### Allow Parameters and RFC compliance Regex
Allow Parameters Regex

### Non Parameters match Regex
Intermediate Regex

Acconut: Parameters in `Content-Type` are also relevant for some media files. Although I am not sure if necessary, they can contain information about the used codecs in audio/video files (see https://github.com/tus/tusd/issues/1194).

Tusd currently does not support parameters, but I plan an changing this. Its implementation currently also uses a regular expression for checking the media type's validity, but my preferred solution is to replace it with [Go's builtin media type parser](https://pkg.go.dev/mime#ParseMediaType). I'm not sure if such a method is easily available to you, but maybe it provides another perspective on this topic.

Murderlon (approved):
Great contribution, thanks!

packages/server/src/handlers/GetHandler.ts
Murderlon: @jesusgoku wow thanks for taking the time to actually test it. Sounds good to me.

PR #666, opened 2024-10-23
Improved redis kv stores

vktrl:
#### Use `SCAN` instead of `KEYS` when getting ids from Redis KV store.

From [redis docs](https://redis.io/docs/latest/commands/keys/):

>Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using [SCAN](https://redis.io/docs/latest/commands/scan/) or [sets](https://redis.io/docs/latest/develop/data-types/sets/).

#### Add support for ioredis

Added a new IoRedisKvStore. `SCAN` and raw command method signatures aren't compatible between node-redis and ioredis. Separate class is more maintainable and safer than making the redis client type an union and trying to figure out which client it's using at runtime.

### DIFF
--- a/.changeset/breezy-llamas-juggle.md
+++ b/.changeset/breezy-llamas-juggle.md
@@ -1,0 +1,5 @@
+---
+"@tus/utils": minor
+---
+
+Add IoRedisKvStore & use redis.scan instead of discouraged redis.keys
--- a/.changeset/smooth-gifts-beam.md
+++ b/.changeset/smooth-gifts-beam.md
@@ -1,0 +1,5 @@
+---
+"@tus/server": minor
+---
+
+Add ioredis as optional dependency
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@tus/file-store": "^1.5.0",
         "@tus/gcs-store": "^1.4.0",
         "@tus/s3-store": "^1.6.0",
-        "@tus/server": "^1.8.0",
+        "@tus/server": "^1.9.0",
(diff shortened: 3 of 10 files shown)


vktrl:
> Code LGTM but we don't have automated tests for this. Can you confirm this works as expected locally?

Yes. I wasn't familiar with how your tests are set up, but I ran some local tests to make sure. Should be easy enough to adapt.

```ts
import {createClient} from '@redis/client'
import {RedisKvStore} from './RedisKvStore'
import {IoRedisKvStore} from './IoRedisKvStore'
import {Redis} from 'ioredis'
import {randomUUID} from 'node:crypto'

const ioRedis = new Redis()
const redis = createClient()

const prefix = `kvstore-test:${Date.now()}:`

const store = new RedisKvStore(redis, prefix)
const ioStore = new IoRedisKvStore(ioRedis, prefix)

const run = async () => {
  const values: string[] = []

  for (let i = 0; i < 25; i++) {
    const value = randomUUID()
    values.push(value)
    await redis.set(`${prefix}${value}`, value, {EX: 60})
  }
  const ioList = await ioStore.list()
  const list = await store.list()
  console.log(list.length === values.length)
  console.log(ioList.length === values.length)
}

redis.connect().then(() => {
  run().then(() => {
    redis.disconnect().then(() => {
      process.exit(0)
    })
  })
})
```

Murderlon (approved):

PR #698, opened 2025-01-16
@tus/s3-store: Change private modifier into protected

Vija02:
Changes the `private` modifier in S3 store into `protected` so that they can be extended. I've also exported the `Options` type so that it can be used while extending.


Murderlon (approved):
Makes sense! Thanks

PR #699, opened 2025-01-17
fix: handling consistent cancellation across stream and locks

### LINKED ISSUE #697: Lock on file never unlocked when request ends prematurely
### Initial checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Forum](https://community.transloadit.com/c/tus/6)

### Steps to reproduce

When a request ends prematurely (eg: when the user refresh the page), it could lead to the file never being unlocked and the promise hanging indefinitely. This only happens when the action occurs on specific timings.

**Steps:**
- Extend the `getUpload` function and artificially delay it for a few seconds.
- Start an upload to the server
- In the middle of the artificial delay, interrupt the request (eg: by refreshing the page)

What's happening is that in https://github.com/tus/tus-node-server/blob/7182c7b75dbf7475dd2fe0e7468cbf39f658f0ed/packages/server/src/handlers/BaseHandler.ts#L181
The `req` is handed over to the pipeline which eventually resolves the Promise. However, if the request is already closed by this point, it never resolves nor rejects. I think adding a quick `req.closed` here will solve the issue?

Separately, I also encountered an issue where if the resource is locked (like this issue) and the user tries to upload, the server will go into an infinite loop and eventually run out of heap memory. Fixing this issue should avoid that entirely but I thought it might be worth looking into as well (from a cursory look, it seems to be looping on trying to acquire the lock)

Not sure what the best fix here is (maybe just reject if it's closed?). If you give me a direction, I could create a PR 

### Expected behavior

It should handle the scenario when the request is over before it gets to that point

### Actual behavior

The whole workflow stayed pending, leaving the lock on the file to remain forever

fenos:
This PR fixes #697 by always cancelling the context when the `req.on('close')` is emitted.

> The close event on the request (req) object is emitted when the underlying connection is terminated before the request is fully processed. This typically happens when the client abruptly disconnects, closes the browser, or cancels the request (e.g., a page refresh or navigation away).

Handled Cancellation scenarios:

- Client aborts the request (signal is propagated to the proxy stream and lockers)
  - If the locker is trying to acquire a lock, it will stop since there is no more need
  - Lock is acquired, the proxy stream is closed gracefully so that we can finishing uploading the remaining bytes in the buffer before releasing the lock
  
- Client completes the request correctly (without keep alive) the on('close') is still triggered, it will not have any effect but probably good practice to make sure all future resources are clean up.

- Client completes the request correctly (with keep-alive) the on('close') is not triggered, context is still cancelled upon req.on('finish'), it will not have any effect but probably good practice to make sure all future resources are clean up.

### DIFF
--- a/.changeset/calm-adults-roll.md
+++ b/.changeset/calm-adults-roll.md
@@ -1,0 +1,6 @@
+---
+"@tus/server": patch
+"@tus/utils": patch
+---
+
+Consistent cancellation across streams and locks, fixing lock on file never being unlocked when the request ends prematurely.
--- a/packages/server/src/handlers/BaseHandler.ts
+++ b/packages/server/src/handlers/BaseHandler.ts
@@ -1,6 +1,6 @@
 import EventEmitter from 'node:events'
 import stream from 'node:stream/promises'
-import {addAbortSignal, PassThrough} from 'node:stream'
+import {PassThrough, Readable} from 'node:stream'
 import type http from 'node:http'
 
 import type {ServerOptions} from '../types'
@@ -121,7 +121,7 @@ export class BaseHandler extends EventEmitter {
 
     const lock = locker.newLock(id)
 
-    await lock.lock(() => {
+    await lock.lock(context.signal, () => {
       context.cancel()
     })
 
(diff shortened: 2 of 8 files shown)


fenos:
@Murderlon when you have sometime please have a test with this.
From my testing seems to work nicely

Murderlon (approved):
LGTM. Thanks for picking it up immediately.

PR #703, opened 2025-01-20
@tus/s3-store: add `minPartSize` option

### LINKED ISSUE #632: @tus/s3-store incompatibility with R2
R2 has a non-standard requirement that all non-trailing parts of a multi-part upload be exactly the same size. With @tus/s3-store@1.4.3 and @tus/server@1.6.0 get the occasional error `InvalidPart: All non-trailing parts must have the same length.` from R2 and the multi-part upload fails.

This is definitely a shortcoming with R2 and not a bug with @tus/s3-store, but we're trying to work around the issue anyhow. Any ideas why @tus/s3-store might generate parts of varying sizes (not counting the final trailing part)?

Thanks!

Murderlon:
Closes #632
Ref: https://github.com/tus/tusd/issues/1137

### DIFF
--- a/.changeset/ten-buttons-suffer.md
+++ b/.changeset/ten-buttons-suffer.md
@@ -1,5 +1,5 @@
 ---
-"@tus/s3-store": patch
+"@tus/s3-store": minor
 ---
 
 Add `minPartSize` option. This can be used alongside `partSize` to guarantee that all non-trailing parts are _exactly_ the same size, which is required for Cloudflare R2.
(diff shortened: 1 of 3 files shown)


.changeset/ten-buttons-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tus/s3-store": patch
```
Murderlon, line 2:
```suggestion
"@tus/s3-store": minor
```

PR #704, opened 2025-01-20
Replace demo folder with StackBlitz

### LINKED ISSUE #343: Upgrade tus-js-client to 3.x and rewrite the demo
`tus-js-client@3.0.0` removed the demo files from the publish so we need to rewrite the demo.

### LINKED ISSUE #547: Demo for GCS incorrectly uses GCSDataStore
line 8 and 14 of demo/server.js incorrectly use GCSDataStore instead of GCSStore
AAnd the whole API call is wrong the the code fails

My Fix

const {Server, EVENTS} = require('@tus/server')
const {GCSStore} = require('@tus/gcs-store')
const {S3Store} = require('@tus/s3-store')
const {FileStore} = require('@tus/file-store')
const {Storage} = require('@google-cloud/storage')

const stores = {
  GCSDataStore: () => {
    const storage = new Storage( {keyFilename: path.resolve(__dirname, '../keyfile.json')})
    return new GCSStore({
      bucket: storage.bucket('tus-node-server'),
    })
    },
 ....

Murderlon:
Closes #547 
Closes #343


mitjap:
I miss this demo setup. This was great way to debug and test various datastores, configurations, isolate issues, reproduce bugs, ...

Maybe I'm not familiar with all tools used to run/debug this project. How do you run `tus-node-server` to debug with real client request and real datastore?

Murderlon:
I personally never used it for that I always start by writing a test if something is wrong which is great because then we also have that going forward. 

If you think it's important, feel free to create an issue for a new dev setup

PR #712, opened 2025-02-05
@tus/s3-store: add `maxMultipartParts` option

### LINKED ISSUE #711: @tus/s3-store: add support to specify `maxMultipartParts`
### Initial checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Forum](https://community.transloadit.com/c/tus/6)

### Problem

Scaleway's S3-compatile Object Storage has different number of parts limit. While AWS S3's limit is `10.000`, Scaleway's is `1.000`. I've reported a bug to them but I don't expect anything to change at their side. 

This is a problem because `@tus/s3-store` calculates required part size for given file size to make sure it does not exceed this limit. 

Server rejects part upload with HTTP status 400 and message `InvalidArgument: Part number must be an integer between 1 and 1000, inclusive`.

AWS S3 documentation:
> Part numbers can be any number from 1 to 10,000, inclusive.
> Soruce: https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html

Scaleway Object Storage documentation:
> The following technical specifications apply to multipart uploads: 1 to 1000 parts per object
> Source: https://www.scaleway.com/en/docs/object-storage/api-cli/multipart-uploads/#uploading-a-part

### Solution

I would like to specify to `@tus/s3-store` what limits are imposed by the server, in this case `maxMultipartParts`.

### Alternatives

I can reduce the impact of this problem by increasing `minPartSize` to a larger value (e.g. `50MB`).

mitjap:
This PR closes #711.

I added additional option `maxMultipartParts` which allows user to define non-standard S3 server limitations.

Also some tests for completeness. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configurable option for multipart uploads, allowing users to set a custom maximum parts limit (defaulting to 10,000) for improved flexibility with S3-compatible storage providers.
  
- **Tests**
  - Expanded the test suite to verify the default and custom behaviors of the new multipart upload configuration.

### DIFF
--- a/.changeset/tricky-emus-fail.md
+++ b/.changeset/tricky-emus-fail.md
@@ -1,0 +1,5 @@
+---
+"@tus/s3-store": minor
+---
+
+Add `maxMultipartParts` option. This can be used when using S3-compatible storage provider with different part number limitations.
--- a/packages/s3-store/src/index.ts
+++ b/packages/s3-store/src/index.ts
@@ -37,6 +37,10 @@ export type Options = {
    * Can not be lower than 5MiB or more than 5GiB.
    */
   minPartSize?: number
+  /**
+   * The maximum number of parts allowed in a multipart upload. Defaults to 10,000.
+   */
+  maxMultipartParts?: number
   useTags?: boolean
   maxConcurrentPartUploads?: number
   cache?: KvStore<MetadataValue>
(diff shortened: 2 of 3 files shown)


Murderlon:
> Are test set to fail if it encounters focused tests? (using `.only()` keyword)?

The linting step (Biome) should fail on this yes

PR #724, opened 2025-02-12
Rename POST_RECEIVE_V2 to POST_RECEIVE

Murderlon:
Do not merge before #710 

`POST_RECEIVE` is called every time an upload finished writing to the store. This event is emitted whenever the request handling is completed (which is the same as `onUploadFinish`, almost the same as `POST_FINISH`), whereas the `POST_RECEIVE_V2` event is emitted _while_ the request is being handled.

`POST_RECEIVE` has been marked deprecated for a long time.


mitjap:
It would be nice to use a different event name for all who are still using deprecated event name. Or is it OK for user to silently transition to use new event?

PR #727, opened 2025-02-14
@tus/s3-store: add `s3Client` option

mitjap:
With this change, S3 Client can be injected instead of being created internally, which enables mocking and improves testability in general.

### DIFF
--- a/.changeset/shy-buckets-knock.md
+++ b/.changeset/shy-buckets-knock.md
@@ -1,0 +1,5 @@
+---
+"@tus/s3-store": minor
+---
+
+Add `s3Client` option. This allows user to provide existing S3 client to datastore.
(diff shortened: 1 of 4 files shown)


Murderlon (changes requested):
I'm not a fan of this change in its current form. I think it's confusing to have `bucket`, `s3ClientConfig`, _and_ `s3Client`. Furthermore, people will get warnings that they import the s3 client without having it installed directly. 

I'm okay with one of the following:
1. Make a breaking change that moves `@aws-sdk/client-s3` to peer dependencies and remove `bucket` and `s3ClientConfig` options. Only keep the new `s3Client`.
2. Leave it as is. No one asked for it (yet) and I haven't run into a scenario where this was blocking to mock a test. All s3 client operations are wrapped in a class method so you can just mock the class method instead.

mitjap:
I'm fine with leaving it as is. It's a good suggestion to just mock a wrapped class method :)

Issue #722, opened 2025-02-10
@tus/s3-store: failed part upload crashes entire server


mitjap:
When `uploadPart` or `uploadIncompletePart` throws (is rejected) entire server crashes. This happens often with Scaleway Object Storage as their service fails with error 500 with non-informative console message `An error was encountered in a non-retryable streaming request.` from AWS library. After listening for `unhandledRejection` event I got some more information.

```
Unhandled Rejection at: Promise {
  <rejected> S3ServiceException [InternalError]: We encountered an internal error. Please try again.
      at throwDefaultError (/app/node_modules/@smithy/smithy-client/dist-cjs/index.js:867:20)
      at /app/node_modules/@smithy/smithy-client/dist-cjs/index.js:876:5
      at de_CommandError (/app/node_modules/@aws-sdk/client-s3/dist-cjs/index.js:4965:14)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async /app/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20
      at async /app/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:483:18
      at async /app/node_modules/@smithy/middleware-retry/dist-cjs/index.js:321:38
      at async /app/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/index.js:315:18
      at async /app/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:109:22
      at async /app/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:136:14 {
    '$fault': 'client',
    '$metadata': {
      httpStatusCode: 500,
      requestId: 'txg5c09a7e2fbe14ba89375-0067aa0421',
      extendedRequestId: 'txg5c09a7e2fbe14ba89375-0067aa0421',
      cfId: undefined
    },
    Code: 'InternalError',
    RequestId: 'txg5c09a7e2fbe14ba89375-0067aa0421',
    HostId: 'txg5c09a7e2fbe14ba89375-0067aa0421',
    Resource: '<s3-object-key>.bin'
  },

```

I believe the problem is that when a `deferred` promise is created it does not handle rejections. It is only inserted into a list.
https://github.com/tus/tus-node-server/blob/81eb03a82658b2e5e97abdb05c26418192320724/packages/s3-store/src/index.ts#L421
[...]

PR #728, opened 2025-02-25
@tus/s3-store: fix unhandled promise rejection

Murderlon:
Ref: #722 

May need another breaking change to close the issue but this is already an improvement for the 1.x line. 

Without these changes, the new test would fail as 3 `uploadParts` attempts were made instead of 2 (stopping at failure)

- [ ] Look into promise rejection warning

```
  S3DataStore
    ✔ should handle multiple concurrent part uploads with some failures (1069ms)
(node:14818) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 12)
```

### DIFF
--- a/packages/s3-store/package.json
+++ b/packages/s3-store/package.json
@@ -16,6 +16,7 @@
   ],
   "scripts": {
     "build": "tsc --build",
+    "pretest": "tsc --build",
     "test": "mocha --timeout 40000 --exit --extension ts --require ts-node/register"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


mitjap:
I also have fix for this. I will create draft PR so we can compare solutions.

Murderlon:
> I also have fix for this. I will create draft PR so we can compare solutions.

Sounds good, this is not done yet anyway

Murderlon:
@mitjap I want to release the major version soon. Should I take this over the finish line, do a release, and then a major release or are you showing your version soon?

Issue #751, opened 2025-04-10
Value of "Upload-Offset" header should be string


hoiekim:
### Initial checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Forum](https://community.transloadit.com/c/tus/6)
- [x] I searched [issues](https://github.com/tus/tus-node-server/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Steps to reproduce

It must happen in all environments and any setup using tus-node-server. Inspect the response headers and take a look at the values, especially "Upload-Offset" header.

### Expected behavior

All header fields and values must be string according to [HTTP spec](https://datatracker.ietf.org/doc/html/rfc7230?utm_source=chatgpt.com#section-3.2.6). Clients must be able to assume so and parse "upload-offset" header accordingly.

### Actual behavior

Some clients automatically convert numbers to string, for example, javascript. So I didn't notice this while writing unit testing in the server side.
```
{
  "access-control-allow-origin": "*",
  "access-control-expose-headers": "Authorization, Content-Type, Location, Tus-Extension, Tus-Max-Size, Tus-Resumable, Tus-Version, Upload-Concat, Upload-Defer-Length, Upload-Length, Upload-Metadata, Upload-Offset, X-HTTP-Method-Override, X-Requested-With, X-Forwarded-Host, X-Forwarded-Proto, Forwarded",
  "date": "Thu, 10 Apr 2025 13:31:45 GMT",
  "content-length": "0",
  "x-powered-by": "Express",
  "tus-resumable": "1.0.0",
  "upload-offset": "3"
}
```

However some other clients ignores non-string headers, for example, Swift(and [TUS client for Swift](https://github.com/tus/TUSKit?tab=readme-ov-file)).
```
{
    "Access-Control-Allow-Origin" = "*";
    "Access-Control-Expose-Headers" = "Authorization, Content-Type, Location, Tus-Extension, Tus-Max-Size, Tus-Resumable, Tus-Version, Upload-Concat, Upload-Defer-Length, Upload-Length, Upload-Metadata, Upload-Offset, X-HTTP-Method-Override, X-Requested-With, X-Forwarded-Host, X-Forwarded-Proto, Forwarded";
    "Content-Length" = 0;
    Date = "Thu, 10 Apr 2025 13:18:24 GMT";
    "Tus-Resumable" = "1.0.0";
    "Upload-Offset" = "";
    "X-Powered-By" = Express;
}
```

PR #752, opened 2025-04-10
@tus/server: make upload-offset a string

hoiekim:
Resolves https://github.com/tus/tus-node-server/issues/751 by simply converting the type of "Upload-Offset" header value from number to string.

### DIFF
--- a/.changeset/happy-countries-ring.md
+++ b/.changeset/happy-countries-ring.md
@@ -1,0 +1,5 @@
+---
+"@tus/server": minor
+---
+
+fix bug: upload-offset header should be string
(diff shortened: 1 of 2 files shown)


.changeset/happy-countries-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tus/server": minor
+---
+
+fix bug: upload-offset header should be string
```
Murderlon, line 5:
```suggestion
"@tus/server": patch
---

Correctly return upload-offset header as a string instead of number
```

Murderlon (approved):
Thanks

PR #811, opened 2026-02-14
Add other Credentials support for Azure store

### LINKED ISSUE #847: Support DefaultAzureCredential for authenticating with azure blob storage
### Initial checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Forum](https://community.transloadit.com/c/tus/6)

### Problem

Currently azure store only supports [account key](https://github.com/tus/tus-node-server/tree/main/packages/azure-store#optionsaccountkey)

Microsoft recommends usign [managed identities](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview) to authenticate between services

### Solution

For partity with [tusd](https://tus.github.io/tusd/storage-backends/azure-blob-storage/#authentication), it could be worth implementing the [DefaultAzureCredential from the @azure/identity sdk](https://learn.microsoft.com/en-us/javascript/api/@azure/identity/defaultazurecredential?view=azure-node-latest)

### Alternatives

Alternatively there is an [open pr ](https://github.com/tus/tus-node-server/pull/811) which could be picked up and cleaned up

sam-ayo:
This PR includes support for using other azure credentials like [DefaultAzureCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet) instead of using the storage account key which like any key comes with it's security risks


Murderlon:
Taking a step back, it would be great if people can just pass their own credentials thing and we don't need to manually wrap every option. Not sure if that makes sense, haven't looked into it enough.

JeromeGill:
Is this PR abandoned? It would be good to get this in some form

sam-ayo:
> Is this PR abandoned? It would be good to get this in some form

I can fix the conflicts on it. But not sure @Murderlon is aligned on it?

packages/azure-store/src/index.ts
```diff
@@ -44,23 +52,17 @@ export class AzureStore extends DataStore {
...
+    const credential =
+      'credential' in options
+        ? options.credential
+        : new StorageSharedKeyCredential(options.account, options.accountKey)
 
-    this.blobServiceClient = new BlobServiceClient(
-      storageAccountBaseUrl,
-      sharedKeyCredential
-    )
+    this.blobServiceClient = new BlobServiceClient(storageAccountBaseUrl, credential)
```
JeromeGill: Maybe we can allow passing in a client instead of having to call new here...

let the consumer construct it themselves and use credentials however they please

Murderlon (approved):

PR #824, opened 2026-04-20
@tus/s3-store: gracefully handle zero byte uploads

itslenny:
This PR resolves an issue with zero byte uploads on S3 or Minio.

## Error message

- S3: `MalformedXML: The XML you provided was not well-formed or did not validate against our published schema`
- Minio: `InvalidRequest: You must specify at least one part`

## Reproduction

Any upload with zero bytes will reproduce this issue. This is a simple repro using `tus-js-client`

```JavaScript
import * as tus from 'tus-js-client'
const upload = new tus.Upload(Readable.from(Buffer.alloc(0)), {
  endpoint,
  chunkSize: 1024,
  uploadSize: 0,
  retryDelays: [],
  uploadDataDuringCreation: true,
  metadata: {
    bucketName: 'my-bucket',
    objectName: 'some/object/path.ext',
  },
  onError: (error) => {
    console.error("onError", error)
  },
  onSuccess: () => {
    console.log(`onSuccess`)
  },
})

upload.start()
```

## Additional context

Including the `StreamLimiter` transform as is done [here in BaseHandler](https://github.com/tus/tus-node-server/blob/8906f1454f171a4b964bcf7a4d6fa41a0636f2b9/packages/server/src/handlers/BaseHandler.ts#L189) causes zero byte chunks to not be emitted. This means that `chunkFinished` is never called and therefore no part is ever created. This results in the error as described above because you cannot complete a multipart upload that has no parts.

The test suite already has a test for zero byte uploads, but it doesn't use `StreamLimiter` so it succeeds. I've added a mostly identical test that includes `StreamLimiter` which consistently reproduces the issue.

As a fix I simply check if there are zero parts and add an empty part before trying to finalize the upload.

This PR also adds the AWS_ENDPOINT env to tests to allow testing locally against Minio

### DIFF
--- a/packages/s3-store/src/index.ts
+++ b/packages/s3-store/src/index.ts
@@ -1,7 +1,7 @@
 import os from 'node:os'
 import fs, {promises as fsProm} from 'node:fs'
 import stream, {promises as streamProm} from 'node:stream'
-import type {Readable} from 'node:stream'
+import {Readable} from 'node:stream'
 
 import type AWS from '@aws-sdk/client-s3'
 import {NoSuchKey, NotFound, S3, type S3ClientConfig} from '@aws-sdk/client-s3'
@@ -231,7 +231,7 @@ export class S3Store extends DataStore {
 
   protected async uploadPart(
     metadata: MetadataValue,
-    readStream: fs.ReadStream | Readable,
+    readStream: fs.ReadStream | Readable | Buffer,
     partNumber: number
   ): Promise<string> {
     const data = await this.client.uploadPart({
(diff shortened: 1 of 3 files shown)


ferhatelmas (approved):

packages/s3-store/src/index.ts
```diff
@@ -463,6 +462,22 @@ export class S3Store extends DataStore {
    * This is where S3 concatenates all the uploaded parts.
    */
   protected async finishMultipartUpload(metadata: MetadataValue, parts: Array<AWS.Part>) {
+    // Handle zero-byte uploads - S3 requires at least one part to complete a multipart upload
+    // S3 allows the last part to be 0 bytes, so we upload a single empty part
+    if (parts.length === 0) {
+      const uploadResult = await this.client.uploadPart({
```
Murderlon: Can't you reuse the `uploadPart` method?

itslenny: Yeah, the only issue is that when I try to pass `Readable.from(Buffer.alloc(0))` I get the error `TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "x-amz-decoded-content-length"`

I updated uploadPart to allow passing a Buffer directly and that works correctly.

packages/s3-store/src/test/index.ts
```diff
@@ -278,6 +280,48 @@ describe('S3DataStore', () => {
     }
   })
 
+  it('should upload an empty part when completing zero byte multipart upload', async function () {
```
Murderlon: Since the existing test didn't work as expected, can you just add `StreamLimiter` to that test and remove this new one?

Murderlon (commented):
Thanks for the PR, one more thing

packages/s3-store/src/test/index.ts
itslenny: Done. I also ran the test with / without my fix to ensure it still fails as expected.

Murderlon (approved):
Thanks!

PR #829, opened 2026-05-07
chore(s3-store): replace @shopify/semaphore with async-mutex

Emperiusm:
## Why

`@shopify/semaphore@3.1.0` was archived by Shopify and surfaces as a `Package no longer supported` deprecation on every `npm install` of `@tus/s3-store`. There is no successor release planned, and downstream projects taking dependency hygiene seriously (e.g. via `npm audit` / install-log linting) need a path off it.

[`async-mutex`](https://www.npmjs.com/package/async-mutex) is actively maintained (~17M weekly downloads), MIT-licensed, has zero runtime deps, and exposes a counting `Semaphore` with an equivalent contract.

## What

`packages/s3-store/src/index.ts` is the only consumer. The diff is small:

```diff
-import {Semaphore, type Permit} from '@shopify/semaphore'
+import {Semaphore} from 'async-mutex'
```

API delta this PR adapts to:

| @shopify/semaphore | async-mutex |
|---|---|
| `await sem.acquire()` returns `Permit` | `await sem.acquire()` returns `[value, releaseFn]` |
| `permit.release(): Promise<void>` | `releaseFn(): void` (sync) |

The previous `permit?.release().catch(() => {})` calls were defending against a Promise rejection that `async-mutex.release` does not have (it's a sync fn that only throws on programmer error — over-release of a permit). The `.catch` no-ops collapse to direct `release()` calls.

## Verified

- `tsc --build` is clean for `@tus/s3-store` after the change.
- Local `mocha dist/test/*.js`: the 12 tests that don't need real AWS credentials pass; the 21 `Region is missing` integration failures are pre-existing and unaffected.

## Notes

- Found via dependency-deprecation cleanup in a downstream consumer (Second Opinion). Filing both the issue context and the patch since the change is small enough to ship together.
- No semver-major behavior change in `@tus/s3-store`'s public surface — only the internal concurrency primitive.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1961,14 +1961,6 @@
         }
       }
     },
-    "node_modules/@shopify/semaphore": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/@shopify/semaphore/-/semaphore-3.1.0.tgz",
-      "integrity": "sha512-LxonkiWEu12FbZhuOMhsdocpxCqm7By8C/2U9QgNuEoXUx2iMrlXjJv3p93RwfNC6TrdlNRo17gRer1z1309VQ==",
-      "engines": {
-        "node": ">=18.12.0"
-      }
-    },
     "node_modules/@sinonjs/commons": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-3.0.1.tgz",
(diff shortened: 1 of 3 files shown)


Murderlon:
Perhaps we should just copy the class from `@shopify/semaphore`, it's been working well for us, it's not a moving target the datastructure isn't going to change, and then we own it and reduce dependency problems and security.

Can you update?

Murderlon:
Superseded by #852, which vendors the existing @shopify/semaphore 3.1.0 implementation and removes the dependency without adding a replacement package. Thanks for identifying and proposing the dependency cleanup.

PR #830, opened 2026-05-09
draft: Add SECURITY.md

Mr-DJ:
Added supported versions and vulnerability reporting details.


Mr-DJ:
Please help me update the relevant email address(es), description and expected timeline for response. I would also suggest defining the threat model and indicating what is in or out of scope so as to prevent ~~slop~~ report flooding.

PR #858, opened 2026-07-30
@tus/server: avoid unnecessary POST_RECEIVE throttling

ferhatelmas:
* avoid its overhead when no listeners exist
* cancel pending trailing timers when uploads settle

---

<a href="https://app.devin.ai/review/tus/tus-node-server/pull/858" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/server/src/test/BaseHandler.test.ts
+++ b/packages/server/src/test/BaseHandler.test.ts
@@ -140,15 +140,27 @@ const consumeStoreWrites = (
   store: sinon.SinonStubbedInstance<DataStore>,
   initialOffset = 0
 ) => {
-  let chunkCount = 0
   let offset = initialOffset
-  const chunkWaiters = new Map<number, () => void>()
+  const chunks: Promise<void>[] = []
+  const resolvers: Array<() => void> = []
+
+  const ensure = (index: number) => {
+    while (chunks.length <= index) {
+      chunks.push(
+        new Promise<void>((resolve) => {
+          resolvers.push(resolve)
+        })
+      )
+    }
+  }
 
   store.write.callsFake(async (readable: Readable) => {
+    let index = 0
     for await (const chunk of readable) {
       offset += (chunk as Buffer).byteLength
-      chunkCount++
-      chunkWaiters.get(chunkCount)?.()
+      ensure(index)
+      resolvers[index]?.()
+      index++
     }
 
     return offset
@@ -156,13 +168,8 @@ const consumeStoreWrites = (
 
   return {
     waitForChunk(target: number) {
-      if (chunkCount >= target) {
-        return Promise.resolve()
-      }
-
-      return new Promise<void>((resolve) => {
-        chunkWaiters.set(target, resolve)
-      })
+      ensure(target - 1)
+      return chunks[target - 1]
     },
   }
 }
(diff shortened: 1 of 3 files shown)


Murderlon (commented):
> Note: I handed this review over to Claude (Claude Code) for a deep code-quality pass. I've read it through and I'm posting it as my review.

---

Behavior goals are right, and the fake-timer tests are more deterministic than the wall-clock one in `Server.test.ts:554`. Also worth noting `cancel()` brings us *into line* with the README, which already documents that the trailing event isn't guaranteed. Two structural things I'd like to change first.

### 1. The optional throttle isn't earning its keep

`BaseHandler.ts:187-203`. The ternary saves one closure allocation per request, in a function that already allocates a `PassThrough`, a `Readable`, and a `StreamLimiter`. The cost you're actually removing is the per-chunk `data` listener — so gate only that:

```ts
const postReceive = throttle(/* unchanged */)

// Only pay per-chunk offset accounting when someone is listening.
if (this.listenerCount(EVENTS.POST_RECEIVE) > 0) {
  let tempOffset = upload.offset
  proxy.on('data', (chunk: Buffer) => {
    tempOffset += chunk.byteLength
    postReceive(tempOffset)
  })
}
```

Same perf win, and it drops the ternary, the `| undefined`, the `if (postReceive)` and the `?.` — production diff shrinks to two lines.

### 2. `writeToStore` shouldn't be a manual `Promise` executor

The PR's whole point is cancel-before-settle, and it enforces that by ordering `.finally()` ahead of `.then(resolve, reject)`. That's correct but unguarded — anyone "tidying" it back to `.then().catch().finally()` silently reverts the fix. And there are *two* settle paths: `proxy.on('error')` rejects outside the pipeline chain, so `finally` isn't guaranteed to have run. The window is microtask-scale so nothing leaks in practice, but the invariant is true by accident.

A `try { return await stream.pipeline(...) } catch { /* AbortError mapping */ } finally { postReceive.cancel(); ... }` gives one settle point and makes it a language guarantee. Bonus: the `biome-ignore noAsyncPromiseExecutor` on line 153 is already stale — there's no `await` in that executor. Happy for this to be a follow-up; if so, please comment that the `.finally()` ordering is load-bearing.

### Smaller

- **Late-listener test.** Fine to sample `listenerCount` once, but `does not start progress tracking for a listener added during a write` asserts an implementation artifact as a spec. Drop it, or document the constraint under `POST_RECEIVE` in the README.
- **`createContext`** is the 7th hand-rolled copy (`PatchHandler`, `PostHandler`, `HeadHandler`, `GetHandler`, `DeleteHandler`, `OptionsHandler`). Please extract to `src/test/utils.ts`, and drop the `{context}` wrapper.
- **`consumeStoreWrites`** is a 25-line bespoke waiter registry; a promise-per-chunk would read better.
- **Fake timers** are installed suite-wide in `before()`, which also fakes mocha's own timeout timer. Won't collide at 100ms ticks, but a hang becomes a CI-job hang instead of a 30s failure. `beforeEach`/`afterEach` limits the blast radius.
- **Changeset** should say the user-visible part plainly: consumers no longer get a final trailing `POST_RECEIVE`; use `POST_FINISH` for the terminal offset.

One correctness note in case it comes up: dropping the `data` listener does not affect backpressure — `stream.pipeline` still drives the proxy, and the pipe was already governing flow.

ferhatelmas:
1. It's not the dominant cost but throttle allocates debounce state and nine nested functions. It's still good to prevent it in two lines.
2. Addressed
3. Addressed
4. #863
5. Addressed
6. Wrong for the version used
7. Addressed

ferhatelmas:
@Murderlon shall we get this in?

packages/server/src/test/BaseHandler.test.ts
```diff
@@ -105,3 +109,304 @@ describe('BaseHandler', () => {
...
+        maxFileSize,
+        context
+      )
+    )
+
+    assert.equal(clock.countTimers(), 0)
+    assert.equal(postReceive.called, false)
+  })
+
+  it('rejects when converting the web stream throws synchronously', async () => {
```
Murderlon: we can drop this test I think, not very useful

packages/server/src/test/BaseHandler.test.ts
```diff
@@ -105,3 +109,304 @@ describe('BaseHandler', () => {
...
+
+    context.abort()
+    controller.close()
+    assert.equal(await write, 5)
+
+    assert.equal(clock.countTimers(), 0)
+    assert.equal(postReceive.called, false)
+  })
+
+  it('emits the last observed offset while a write is stalled', async () => {
```
Murderlon: already covered by another test:

 - The stream remains open.
 - A chunk is consumed.
 - The fake clock advances by the interval.
 - The latest accumulated offset is emitted before the write completes.

this only adds

 1. An interval - 1 boundary assertion, which mostly tests lodash’s throttle implementation.
 2. An explicit settled === false assertion, even though the controlled stream being open already establishes that in the other test.

Murderlon (commented):
minor things, other than that LGTM

packages/server/src/test/BaseHandler.test.ts
ferhatelmas: dropped it

packages/server/src/test/BaseHandler.test.ts
ferhatelmas: improved the test to guard promise refactor (`new Promise(async (...) => ...)`) because there is no other test covers it

Murderlon (approved):

PR #877, opened 2026-09-21
@tus/s3-store: avoid false completion on generic 404

ferhatelmas:
A generic `NotFound` from S3 `ListParts` could make `getUpload()` report the full upload offset even when the upload was incomplete. Exclude that error from completion detection while preserving the existing `NoSuchUpload` and `NoSuchKey` behavior.

Cover all three error codes through SDK `name` and legacy `Code` fields, including unchanged propagation of generic errors.

Related to #859


ferhatelmas:
@Murderlon this is a regression fix, it would be nice to get in so that we can bump the versions, thanks 🙏🏻

Murderlon (approved):
Thanks

